PR #5, opened 2026-05-20
fix(audio): restore transmit on system default microphone input

xogium:
System default used AVAudioEngine while named devices used AUHAL with
  explicit Core Audio device binding, which broke channel transmit when
  TeamTalk also held output and the virtual input device.
  - Capture all inputs via AUHAL with an explicit device ID
  - Resample PCM to the channel codec rate before TT_InsertAudioBlock
  - Stop Preferences preview before starting transmit capture
  - Add throttled capture/inject diagnostics and mic-restart status text
  - Log AEC reference channel mismatches; passthrough after repeated AEC drops

### DIFF
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -105,6 +105,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         AudioLogger.clear()
         let sdkVersion = String(cString: TT_GetVersion())
         AudioLogger.log("App launched — TeamTalk SDK %@", sdkVersion)
+        #if DEBUG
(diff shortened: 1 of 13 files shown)


xogium:
I've had to do this on my end in order to allow the system default input option to actually work. If this isn't done, one has to explicitly select the input in the list such as airpods or macbook microphone. It was also causing a lot of problems such as unstability in the audio system that now appear solved, which caused it to restart at random and to fall back on system default input which would once again cause it to be silent. Sometimes it would crash the audio system so bad with airpods for example when they were directly selected that I'd need to completely disconnect the airpods and put them back in again, then go back in audio preferences to set the input to airpods again.

PR #6, opened 2026-05-20
fix(auto-away): stop spurious away toggling and restore on input

xogium:
## Summary

- Fix automatic Away status flipping back to Available within seconds without user input, caused by unreliable IOKit `HIDIdleTime` reads (failures returned 0 and looked like activity).
- Use `CGEventSource` keyboard/mouse idle tracking and require a sharp idle drop from a tracked peak before clearing Away, with a short post-activation grace period to ignore status-change glitches.
- Poll every 0.5s while Away so keypresses restore Available promptly; keep 5s polling when not away.

## Test plan

- [x] Set auto-away timeout to 1–3 minutes, stay idle until Away is applied, confirm status stays Away without touching keyboard/mouse
- [x] Press a key or click the mouse while Away; confirm status returns to Available within ~1s
- [x] Set auto-away timeout to 0; confirm manual Away/Available still works
- [x] Change Away status message manually while auto-away is active; confirm no spurious toggling

### DIFF
--- a/App/ttaccessible/Models/TeamTalkStatusMode.swift
+++ b/App/ttaccessible/Models/TeamTalkStatusMode.swift
@@ -14,6 +14,10 @@ enum TeamTalkStatusMode: Int32, CaseIterable, Equatable {
 
     private static let modeMask: Int32 = 0x000000FF
 
+    static func isAwayStatus(_ bitmask: Int32) -> Bool {
+        (bitmask & modeMask) == away.rawValue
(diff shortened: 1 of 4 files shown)


xogium:
I've just tested this and can confrim the implemented fix actually works out just fine on my end.

Without this fix, auto away would enable itself, then almost immediately get disabled without any action on the user's part, which rendered the feature broken and of not much use.

PR #7, opened 2026-05-21
feat(media): in-channel media streaming, playback fixes, and format support

xogium:
## Summary

- Adds in-channel media file streaming with embedded playback controls (play, pause, seek, broadcast gain) directly in the connected-server view
- Replaces the separate media streaming window with a collapsible in-window video panel so the main UI stays fully usable while playback controls remain within easy reach
- Adds optional system `ffmpeg` pre-conversion via `MediaTranscodeService` when the TeamTalk SDK cannot open a file format (playback itself uses the SDK's bundled FFmpeg)

**Out of scope:** webcam video capture/transmission. The macOS TeamTalk 5 SDK uses 0RGB mode for camera capture, which fails on modern Mac hardware and cameras. Fixing that would require patching the SDK binary — something that feels extremely hackish and is deliberately not included in this PR.

## Testing notes

I did what testing I could given that I am totally blind: keyboard navigation, VoiceOver announcements, control flow, and audio-side behavior. I have **not** been able to visually verify video rendering. Someone with some vision left should double-check the collapsible video panel for rendering glitches, layout issues, or other visual problems.

## Test plan

- [x] Connect to a server and stream a media file the SDK opens natively; confirm audio and video play in the in-window panel
- [x] Collapse and expand the video panel; confirm preference persists across sessions
- [x] Use embedded controls (play, pause, seek, broadcast gain) while navigating the rest of the main UI
- [x] Stream a file the SDK cannot open; confirm optional `ffmpeg` pre-conversion path works when `ffmpeg` is installed
- [x] Confirm remote users receive the stream as expected
- [x] **Visual pass:** verify video frames render correctly, panel sizing/scrolling looks sane, and nothing overlaps or clips oddly
- [x] Confirm no regression to push-to-talk / microphone preferences from the rebase onto `main`


Jookia:
Xogium sent me a video showing the video playing and it visually looks correct and functional.

xogium:
## Update: dropped in-app transcoding, upfront format checks instead

This branch was updated after real-world testing with a ~5.5 GB, ~26 minute 4K HEVC 10-bit phone recording (HEVC Main 10, 3840×2160, yuv420p10le). That file exposed why the “normalize then stream” path is not viable in ttaccessible.

### The problem we hit

1. TeamTalk / phone-style media — Modern phone and camera files are often HEVC (H.265), 4K, and 10-bit HDR. The TeamTalk SDK uses bundled FFmpeg to read a file and streams video to the channel as WebM VP8. The public API does not document “H.264 only”; in practice many of these sources either fail TT_GetMediaFileInfo, fail when streaming starts, or are too heavy to be practical.

2. In-app transcode was unacceptable — We briefly experimented with normalizing unsupported files via system ffmpeg before streaming (full remux/transcode to H.264 at most 1280p). On the test file above, that meant on the order of ~26 minutes of waiting before playback could begin — worse UX than telling the user the file isn’t suitable. For long 4K HEVC sources, that approach is a non-starter.

3. Blocking the connection queue — Running ffprobe / TT_GetMediaFileInfo on huge files on the same serial queue as TeamTalk events could stall the client long enough to drop the server connection, leaving a “still connected” UI while mic/actions failed with “The connection to the server failed.” Preflight now runs off the connection queue; reconnect/stale-session handling was tightened.

### What we did (latest commit)

- Removed MediaTranscodeService and all in-app conversion.
- Added MediaStreamCompatibility: proactive checks before TT_StartStreamingMediaFileToChannelEx.
- Allowed video (when ffprobe is available): H.264, MJPEG (e.g. JPEG-in-AVI), MPEG-4, up to 1280p, 8-bit.
- Blocked upfront: HEVC/H.265, VP9, AV1, 10-bit, and resolution over 1280p — with clear alerts instead of a long transcode or a wedged session.
- README and file picker copy updated so expectations are explicit: convert HEVC/4K/10-bit elsewhere (or use desktop share), don’t expect ttaccessible to transcode them.

### What the SDK vs the app supports

TeamTalk SDK: reads many container/codec combos via FFmpeg; outputs VP8 on the channel for video. No guarantee that 4K HEVC 10-bit phone files work well.

ttaccessible (this PR): streams without conversion. Only starts files that pass our compatibility rules; otherwise shows why (codec, resolution, 10-bit, or SDK can’t open).

### Trying the test file now

That 4K HEVC sample was rejected immediately with a detailed message (codec, resolution, 10-bit), not churn for half an hour. MJPEG AVI and typical H.264 MP4 within 1280p should be allowed through preflight when ffprobe is available (brew install ffmpeg installs it; only used for checking, not transcoding).

Happy to adjust the allowlist further if we find other codecs the SDK handles reliably (e.g. MPEG-2), but we’re intentionally not bringing back full-file transcode in the client.

xogium:
## Follow-up improvements on this branch

Since the original PR description, `stream-improvement` has picked up another round of work (including merging `fix-streams`). Highlights:

### Playback and accessibility
- **Merged `fix-streams`** — pause/seek/resume position fixes (`mediaStreamingSeekedWhilePaused`), one-shot “streaming started” history (no spam on seek/resume), and removal of duplicate VoiceOver announce on API success.
- **`MediaPlaybackPositionControl`** — accessible position readout (elapsed/total as MM:SS) instead of a raw slider that flooded VoiceOver with percentage updates.
- **Position during playback** — thumb and time label keep moving while playing; accessibility value updates silently (no `.valueChanged` spam). VoiceOver reads the current time when you focus the control.

### Format handling
- **Upfront compatibility checks** via `MediaStreamCompatibility` + ffprobe — supported formats are tried without in-app transcoding; unsupported codecs/resolution/10-bit get clear errors before streaming starts.
- **Confirmed streamable video codecs** (up to 1280p on the longest side): **H.264, MJPEG, MPEG-1, MPEG-4**. HEVC, VP9, and AV1 remain blocked.

### Still out of scope
Webcam capture/transmission (SDK 0RGB limitation on modern Mac hardware) — unchanged from the original PR.

---

**Branch status:** This is intended to be the **last batch of improvements** on `stream-improvement` unless reviewers or users want something else. From my side it feels **feature-complete** for in-channel media streaming: embedded controls, collapsible video panel, solid playback fixes, and a clearer format story.

Happy to adjust if anyone sees gaps — otherwise this is ready for a final look (especially visual check of the video panel for sighted reviewers).

math65:
Hey Casey, thanks for the huge amount of work on this — the threading fix (preflight off the connection queue) and the upfront compat checks are great improvements. Did a careful read of the diff, traced what `.rootTree` actually does, and cross-checked against the Qt reference client. Here's what I found.

## 🔴 Potential perf regression on `CLIENTEVENT_USER_STATECHANGE`

In `TeamTalkConnectionController+Connection.swift`, this event now triggers both `publishAudioRuntimeUpdateLocked` and `publishSessionLocked(..., invalidation: .rootTree)`. Tracing what `.rootTree` does in `makeSessionSnapshotLocked`: it sets `requiresRootTree = true`, which re-runs `fetchServerChannelsLocked`, `fetchServerUsersLocked`, and rebuilds the whole tree (channel sort, users-by-channel dict, subscription state map, etc.). Combined with the `fetchServerUsersLocked` that `publishAudioRuntimeUpdateLocked` already does, that's two full user fetches + a full channel fetch + a Swift tree rebuild every time anyone in the server toggles talking/mute.

For reference, the Qt client (`channelstree.cpp:1615`, `ChannelsTree::slotUserStateChange`) handles this event with a single targeted item update — `getUserItem(user.nUserID)` + `slotUpdateTreeWidgetItem`. No full user/channel fetch, no rebuild.

I'm guessing the addition was to propagate the new `isStreamingMediaFileVideo` flag down to `ConnectedServerUser`. If that's the only reason, a much cheaper option would be to add `isStreamingMediaFileVideo` to `ConnectedUserAudioState` (the struct built by `publishAudioRuntimeUpdateLocked`, which already carries `isTalking`, `isMuted`, `isMediaFileMuted`). That pattern already exists in the codebase, propagates per-row, and skips the tree rebuild entirely.

## 🟡 Minor bugs

**Dead ternary in `publishVideoDisplayStateLocked`** (`+Video.swift`):
```swift
frame: clearFrame ? nil : nil
```
The `clearFrame` parameter has no effect — both branches return nil. Looks like a leftover from a refactor.

**Incomplete rollback in `setMediaStreamingPausedLocked`** (`+MediaStreaming.swift`): when `applyMediaStreamingUpdateLocked` fails, the rollback restores `mediaStreamingPaused` and `mediaStreamingUserPauseIntent` but not `mediaStreamingElapsedMSec`, and `mediaStreamingElapsedSampleAt` gets reset to `Date()` rather than its previous value — small elapsed-time drift if the SDK update ever fails mid-pause. Unlikely to fire in practice but the rollback isn't symmetric.

**Private notification name** in `MediaPlaybackPositionControl`:
```swift
forName: Notification.Name("AXFocusedUIElementChanged")
```
This isn't a public API. Works today, but could break silently on future macOS versions. Worth a comment at least, ideally a fallback if the observer never fires.

## ⚠️ Worth a note in the code

**Pixel format assumption** in `VideoFrameView`: the bitmap info assumes BGRA (`byteOrder32Little | noneSkipFirst`). True for SDK-decoded WebM VP8, but undocumented in the file. If the SDK ever changes, we'd get silently broken colors / black frames. A one-line comment referencing the SDK guarantee would help future debugging.

**Frame copy cost**: `copyVideoFramePayload` allocates a `Data` per frame via `Data(bytes:count:)`. At 1280×720 RGBA that's ~3.7 MB per frame; at 30 fps it's ~110 MB/s of copies hopping to the main thread. Probably fine in practice for short streams, but a point worth flagging — if perf testing shows hotspots, the SDK `TT_Acquire`/`TT_Release` pattern allows keeping the buffer until render without copying.

## ✅ Things I really liked

- `shouldIgnoreMediaStreamingFinalizeLocked` with the `mediaStreamingFinalizeSuppressedUntil` window + distance-to-end check — clean way to ignore the parasitic `MFS_FINISHED` events during seeks
- `restartMediaStreamLocked` with the inline comment about `TT_Update` being unreliable for video — exactly the kind of "why" comment that's useful
- `MediaPlaybackPositionControl` accessibility pattern (`accessibilityChildren() -> []`, role slider, valueChanged only on focus/drag/pause) — clean and consistent with `AudioGainControlView`
- Clean removal of `MediaStreamingPlayerWindowController` and all its AppDelegate plumbing
- `SessionGuard` extension is small, focused, and reused consistently across Audio/Connection/MediaStreaming

Nothing here is a blocker — the `.rootTree` invalidation is the one I'd really want to confirm intent on before merging. The rest is cleanup. Happy to take a manual test pass once you've had a look.

_(Review assisted by Claude — happy to dig deeper into any point if I got something wrong.)_

xogium:
## Response to latest code review

Updates pushed on `stream-improvement` since `2415c760c6`. Below is what each commit does and how it addresses review feedback.

---

### `2415c76` — perf(connection): avoid full tree rebuild on user state change

What it does: Stops republishing the entire connected-server tree snapshot on every `CLIENTEVENT_USER_STATECHANGE`.

How: Adds `isStreamingMediaFileVideo` to `ConnectedUserAudioState` so per-user audio/runtime updates can carry the media-file video flag. `ConnectedServerViewController` merges those partial updates into existing user rows (including `isStreamingMediaFileVideo`) instead of replacing the whole `rootTree`. `updateVideoSelectionFromTree()` runs after audio updates so the video panel selection stays in sync when a user starts or stops streaming video.

Why: Review noted unnecessary full tree rebuilds on frequent state-change events; this keeps updates scoped to changed users.

---

### `c8b2658` — fix(video): honor clearFrame in metadata-only display updates

What it does: Fixes the in-window video panel clearing or losing its image when only display metadata changes.

How: Caches `lastPublishedVideoFrame` and `lastPublishedVideoFrameUserID` when a frame is published. `publishVideoDisplayStateLocked(clearFrame:)` reuses the cached frame when `clearFrame` is false and the active user matches; only passes `frame: nil` when `clearFrame` is true or there is no cached frame for that user.

Why: Metadata-only publishes were effectively always sending an empty frame, so the panel could go blank even while streaming continued.

---

### `91207fd` — fix(media-streaming): restore elapsed time on pause update failure

What it does: Keeps the scrubber position consistent if a pause/resume SDK update fails.

How: When `applyMediaStreamingUpdateLocked` fails in `setMediaStreamingPausedLocked`, rolls back `mediaStreamingElapsedMSec` and `mediaStreamingElapsedSampleAt` (in addition to the existing pause-state rollback).

Why: On failure we were restoring pause flags but leaving elapsed time at a value computed mid-update, which could desync the UI from the actual stream.

---

### `7fd4179` — docs(video): document BGRA layout for SDK-decoded frames

What it does: Documents that SDK-decoded video frames are BGRA in `VideoFrameView`.

How: Adds an inline comment at the rendering site so future readers know the byte layout TeamTalk delivers.

Why: Review asked for clarity on pixel format assumptions in the video display path.

---

### `b354811` — docs(video): note per-frame copy cost in copyVideoFramePayload

What it does: Documents performance characteristics of copying each video frame for Swift/UI use.

How: Comments in `TeamTalkConnectionController+Video.swift` on `copyVideoFramePayload` explaining that each frame is copied and that cost is intentional for thread-safe UI handoff.

Why: Review requested acknowledgment of per-frame allocation/copy cost in the hot path.

---

### `cef57da` — docs(a11y): note private AXFocusedUIElementChanged notification

What it does: Documents a non-public accessibility API used by the media position scrubber.

How: Adds a comment in `MediaPlaybackPositionControl` where we observe `AXFocusedUIElementChanged` to detect VoiceOver focus without spamming announcements during playback.

Why: Review asked why a private AppKit notification is used; the comment explains the VoiceOver focus pattern and that it is not in the public API.

---

### `46e6dba` — fix(media-streaming): restore playback position after pause, seek, and resume

What it does: Fixes incorrect playback position after pause → seek (scrubber) → resume for both audio and video streams.

Problem (video): After `2a7cbb6`, video seeks used stream restart, but resume still used `TT_Update` with `TT_MEDIAPLAYBACK_OFFSET_IGNORE`. The UI showed the scrubbed time; the SDK often restarted from 00:00. The `5073d62` `mediaStreamingSeekedWhilePaused` path was never wired for video after the merge at `a1ad2007`.

Problem (audio): Seeks/resume relied on `TT_Update`; the SDK frequently kept position near the old pause point (~5–10 s behind the scrubber). `CLIENTEVENT_STREAM_MEDIAFILE` progress events could overwrite app elapsed time before playback caught up.

How:
- All seeks: `restartMediaStreamLocked` at the target offset (stop + `StartStreamingMediaFileToChannelEx`) instead of `TT_Update` to reposition.
- Resume after paused seek: same restart at scrubber offset for audio and video (not `IGNORE` / not update-only).
- While paused after seek: ignore SDK elapsed updates so `MFS_*` events do not reset the scrubber before resume.
- After a playing restart: 2 s resume anchor ignores regressive SDK elapsed (e.g. brief 0 ms after `MFS_STARTED`) so the UI does not jump backward.

Plain pause/resume without a seek in between still uses `TT_Update` + `IGNORE`, which is correct when the SDK position is already aligned.

---

Happy to adjust anything if a specific review thread should be called out by name. Thanks again for the thorough pass.

math65 (approved):
Approved. The 7 fixes from the last review round all landed cleanly: the perf regression on `CLIENTEVENT_USER_STATECHANGE` is gone (per-row `isStreamingMediaFileVideo` propagation works exactly as suggested), the `clearFrame` dead ternary is replaced by a proper cached-frame path, the pause-rollback is now symmetric, and the docs comments on BGRA/copy cost/AXFocusedUIElementChanged are clear. The bonus `d8e9ce2` (defer codec/resolution to SDK) is sensible — the SDK does decode 8-bit HEVC on macOS, the upfront ffprobe gates were too aggressive.\n\nManually tested: streaming, pause→seek→resume, user state changes while streaming. All good.\n\nMinor follow-ups (not blocking):\n- The FR `mediaStream.panel.message` reads as a slight calque of the English ("La prise en charge vidéo dépend de ce que TeamTalk peut ouvrir sur votre Mac" — more natural would be "…dépend des formats que TeamTalk peut décoder sur votre Mac")\n- After `d8e9ce2`, the strings `mediaStream.error.reason.videoCodec`, `videoCodecBlocked`, and `videoResolution` are no longer referenced and can be removed (EN + FR)\n- Tiny nit in `updateMediaStreamingProgressLocked`: `if mediaStreamingResumeAnchorMSec != nil, let anchorMSec = mediaStreamingResumeAnchorMSec` — the `!= nil` is redundant with the `let`\n\nHappy to polish these post-merge. Thanks for the thorough round of fixes.

PR #8, opened 2026-05-21
Improve nickname UX: derive from computer name, apply live on change

Quinton1110:
## Summary

- **Smart default nickname**: parsed from the Mac's computer name — e.g. \"Quinton's MacBook Air\" → \"Quinton\" — so most users connect with a sensible name without any setup
- **Live nickname updates**: changing the nickname in Preferences now takes effect immediately on any connected server, no restart required
- **Keychain error recovery**: when a saved password can't be read from the keychain (e.g. after re-signing), the app now opens the credential editor instead of showing a dead-end error alert

## Test plan

- [ ] Fresh install (no saved preferences): verify nickname defaults to first name from computer name
- [ ] Change nickname in Preferences while connected: verify it updates on the server immediately
- [ ] Simulate keychain failure: verify \"Edit Credentials\" dialog appears instead of a dead-end error


Quinton1110:
Closing in favour of a cleaner PR with only the relevant changes.

PR #9, opened 2026-05-21
Improve nickname UX: derive from computer name, apply live on change

Quinton1110:
## Summary

- **Smart default nickname**: parsed from the Mac's computer name — e.g. \"Quinton's MacBook Air\" → \"Quinton\" — so most users connect with a sensible name without any setup
- **Live nickname updates**: changing the nickname in Preferences now takes effect immediately on any connected server, no restart required
- **Keychain error recovery**: when a saved password can't be read from the keychain (e.g. after re-signing), the app now opens the credential editor instead of a dead-end error alert

## Test plan

- [ ] Fresh install (no saved preferences): verify nickname defaults to first name from computer name
- [ ] Change nickname in Preferences while connected: verify it updates on the server immediately
- [ ] Simulate keychain failure: verify \"Edit Credentials\" dialog appears instead of a dead-end error

### DIFF
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -104,7 +104,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         userDriverDelegate: nil
     )
     private var updaterAutoCheckCancellable: AnyCancellable?
-
+    private var nicknameCancellable: AnyCancellable?
     private var pushToTalkModeCancellable: AnyCancellable?
 
     func applicationDidFinishLaunching(_ notification: Notification) {
@@ -140,6 +140,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         handleLaunchTTFilesIfNeeded()
         processPendingTTFileURLsIfPossible()
         syncSparkleAutoCheckPreference()
+        syncNicknamePreference()
         scheduleLaunchUpdateCheck()
         configurePushToTalkObservers()
     }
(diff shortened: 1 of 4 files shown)


math65:
Hey Quinton, thanks for the PR — the idea of pre-filling the nickname is really nice.

I tested on my machine and wanted to flag a case the current logic doesn't cover: on a Mac set up in French, the default computer name doesn't follow the "Quinton's MacBook Air" pattern but rather "MacBook Pro de Mathieu". No apostrophe, so we fall through to the `firstWord` fallback, which returns "MacBook" as the nickname. Same for Spanish ("MacBook de Juan"), Portuguese, Italian ("di"), etc. — all the Romance languages put the first name at the end.

An alternative that might be more robust: instead of parsing the computer name, read the full account name directly via `NSFullUserName()`. On my Mac it returns "Mathieu Martin"; on yours it's probably "Quinton Williams". Then just take the first word, and it works across all languages without having to maintain per-locale patterns.

Edge cases I can think of:
- Account with no full name (rare) → `NSFullUserName()` returns the short name (e.g. "math65") — not ideal, but at least it's something macOS actually knows about the user
- Compound first names ("Jean-Pierre Dupont") → "Jean-Pierre", fine
- Hispanic names where the order is reversed — edge case

What do you think? Happy to make the change myself, or you can do it — whichever you prefer.

Quinton1110:
Hey, I'm so sorry I only just now saw this email.
I suppose I should have thought it more through, but I'm also  monnolingual.
Either way, I do think it  could be helpful to have a way of prefilling the name. I did see an update was out today but  still have to see what was changed in the repo.

math65:
Hey Quinton, thanks again for putting this together — I've integrated your changes into main as commit 2945ba1, with you credited as co-author. Closing this PR since the work is now landed directly.

A couple of small adjustments on top of your branch:

1. **Nickname derivation** — replaced the computer-name heuristic with `NSFullUserName()` + first word, which works across all locales (Romance languages like French, Spanish, Portuguese put the first name at the end of the computer name, so "MacBook Pro de Mathieu" was falling through to "MacBook"). Falls back to `"TTAccessible"` when the account has no full name set.

2. **`openCredentialEditor` refactor** — kept your nice extraction and applied it to `connectSelectedServer`. The `editSelectedServer` path got a related fix in 1.3.1 (uses `try?` to swallow the keychain error and open the editor with empty fields directly), so I left that one alone to avoid stacking two rescue flows on top of each other.

3. **Live nickname sync** — `syncNicknamePreference()` in AppDelegate is exactly your code.

It'll ship in the next release (1.3.2-ish, alongside the Casey PR #5 review TODOs). Cheers for the contribution!

PR #10, opened 2026-05-23
Fix build docs for contributors without signing certs

xogium:
## Summary

- Document `CODE_SIGNING_ALLOWED=NO` for local Debug and Release `xcodebuild` commands, since the project is configured for automatic signing with the maintainer's Apple Development certificate.
- Correct the `build.sh` section: it does not produce an unsigned app for other contributors and remains maintainer-only, with clearer notes on `--notarize` and `--release`.

## Test plan

- [x] Follow the updated Debug `xcodebuild` command on a machine without the maintainer signing certificate and confirm the build succeeds.
- [x] Read the Packaging section and confirm it matches how `build.sh` behaves for maintainers vs contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,21 +52,27 @@ This downloads `libTeamTalk5.dylib` and `TeamTalk.h` from the [official TeamTalk
 
 For development:
 
+The Xcode project uses automatic code signing with an **Apple Development** certificate tied to the maintainer's team. If you do not have that certificate, disable code signing when building locally:
+
 ```bash
 # Debug build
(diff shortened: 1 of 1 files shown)


math65 (approved):
Thanks Casey — good catch. The old README was misleading: `build.sh` does require the Apple Development cert (no `CODE_SIGNING_ALLOWED=NO` in the xcodebuild invocation), so contributors without it would have hit an error. The new wording is accurate, and adding `CODE_SIGNING_ALLOWED=NO` to the Debug/Release examples is exactly the right move for local builds.

PR #11, opened 2026-05-24
Polish post-merge media streaming follow-ups

xogium:
## Summary
- Improve French `mediaStream.panel.message` copy: "dépend des formats que TeamTalk peut décoder" instead of the calque "dépend de ce que TeamTalk peut ouvrir"
- Remove unused `mediaStream.error.reason.videoCodec`, `videoCodecBlocked`, and `videoResolution` strings (EN + FR) — no longer referenced after deferring codec/resolution checks to the SDK (d8e9ce2)
- Simplify `updateMediaStreamingProgressLocked`: drop redundant `!= nil` before `if let anchorMSec = mediaStreamingResumeAnchorMSec`

## Test plan
- [ ] Open "Stream Media File" panel and confirm FR message reads naturally
- [x] Confirm unsupported-format errors still show `sdkUnsupported` / `tenBitVideo` where applicable
- [x] Smoke-test pause → seek → resume on a media stream (resume anchor clearing unchanged in behavior)

### DIFF
--- a/App/ttaccessible/Services/TeamTalkConnectionController+MediaStreaming.swift
+++ b/App/ttaccessible/Services/TeamTalkConnectionController+MediaStreaming.swift
@@ -398,8 +398,7 @@ extension TeamTalkConnectionController {
             publishMediaStreamingProgressLocked()
             return
         }
-        if mediaStreamingResumeAnchorMSec != nil,
-           let anchorMSec = mediaStreamingResumeAnchorMSec,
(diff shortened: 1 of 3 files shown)


math65 (approved):
Perfect — exactly the three items from the previous review, nothing more. The FR rewording matches the suggestion verbatim, the dead strings are gone in both locales, and the redundant nil check is cleaned up. Thanks for the quick turnaround.

Issue #1, opened 2026-04-13
Feature request: allow for muting media file stream


johanntan:
I know you can already use `ctrl+6` to unsubscribe from media file stream, but can you add the `cmd+ctrl+shift+m` to mute/unmute it, similar to `cmd+shift+m` on voice?
Also, the ability to adjust media file stream volume in the `cmd+u` dialog would be quite nice to.

math65:
Both implemented and shipped in 1.0.0-beta.12:

- **Cmd+Ctrl+Shift+M** mutes/unmutes the selected user's media file stream (paired with Cmd+Shift+M for voice). Wired in `ttaccessibleApp.swift` and handled by `toggleMuteUserMediaFileAction`.
- The **Cmd+U** dialog exposes a separate slider for media file volume in addition to the voice volume slider.

Release: https://github.com/math65/ttaccessible/releases/tag/v1.0.0-beta.12

PR #14, opened 2026-05-30
fix(audio): restart only on meaningful route changes

xogium:
## Summary

Fixes unsolicited full audio restarts that caused ~1–2 second cutouts on both input and output, and could freeze the app when CoreAudio fired device-list notifications frequently.

**Problem:** On every `kAudioHardwarePropertyDevices` event (and related CoreAudio notifications), `AppDelegate` debounced a call to `TT_RestartSoundSystem()`. That tore down the mic, virtual input, output, and AEC speaker tap, then reopened everything—often ~2 seconds of dead air. Benign churn (Continuity “iPhone Microphone” appearing/disappearing, virtual cables, the app’s own AEC aggregate device on start/stop) triggered the same path as a real hardware change. A restart could also schedule another restart when the tap was destroyed, producing back-to-back outages.

**Fix:** Route hardware changes through a debounced handler that compares an `AudioRoutingSnapshot` (resolved input, system default in/out UIDs, preferred output still in catalog, active mic sample rate). Only restart when something that affects the live session actually changed:

- System default input or output changed (e.g. AirPods become default)
- Explicitly chosen input/output device unplugged or removed from the catalog
- Nominal sample rate changed on the active input device

Otherwise refresh the TeamTalk device cache only (`catalog refresh only` in logs)—no `TT_RestartSoundSystem`.

Additional behavior:

- **Preferences device changes** while audio is active still use a full `restartSoundSystem` so TeamTalk picks up the new routing reliably.
- **Suppression** during mic enable, manual restart, and speaker-tap lifecycle avoids restart cascades (including the first-time microphone permission dialog backlog).
- **Monitor** passes the CoreAudio property selector in the notification for sample-rate-specific handling.

## Evidence

Before: logs showed dozens of `restartSoundSystem: begin` per session, often in pairs within seconds, correlated with `kAudioHardwarePropertyDevices` and device count toggling 8↔9 (Continuity).

After: typical session shows `catalog refresh only` for irrelevant list changes, single `restarting sound system for route change` when system defaults change (e.g. AirPods connect/disconnect), aligned `inserts`/`chunks` on first mic enable, no restart storm on quit.

## Test plan

- [x] Connect to a server, enable mic (grant permission on first use)—no long delayed/echoed burst after allowing access
- [x] Use audio for several minutes with Continuity disabled—no random ~2 s cutouts; log should not spam `restartSoundSystem: begin`
- [ ] With a fixed (non-default) mic selected, plug/unplug unrelated devices—audio stays up; log shows `catalog refresh only`
- [x] Set input/output to System Default, connect/disconnect AirPods—one restart per real route change; audio follows defaults
- [ ] Change input/output in Preferences while connected—audio moves to selected devices
- [x] Manual “Refresh devices” in Preferences still performs a full restart

### DIFF
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -94,7 +94,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var pendingUnsavedServerConfiguration: PendingUnsavedServerConfiguration?
 
     private var deviceChangeObserver: Any?
-    private var globalDeviceChangeWorkItem: DispatchWorkItem?
 
     private lazy var updaterController: SPUStandardUpdaterController = SPUStandardUpdaterController(
         startingUpdater: true,
@@ -121,13 +120,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             forName: AudioDeviceChangeMonitor.audioDevicesDidChange,
             object: nil,
             queue: .main
-        ) { [weak self] _ in
-            self?.globalDeviceChangeWorkItem?.cancel()
-            let workItem = DispatchWorkItem { [weak self] in
-                self?.connectionController.restartSoundSystem { _ in }
-            }
-            self?.globalDeviceChangeWorkItem = workItem
-            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500), execute: workItem)
+        ) { [weak self] notification in
+            let selector = notification.userInfo?[AudioDeviceChangeMonitor.selectorUserInfoKey] as? UInt32 ?? 0
+            self?.connectionController.handleDebouncedAudioHardwareChange(selector: selector)
         }
         requestNotificationPermission()
         showSavedServersWindow()
--- a/App/ttaccessible/Models/AudioRoutingSnapshot.swift
+++ b/App/ttaccessible/Models/AudioRoutingSnapshot.swift
@@ -1,0 +1,18 @@
+//
+//  AudioRoutingSnapshot.swift
+//  ttaccessible
+//
+
+import Foundation
+
+/// Captures the audio routing state that matters for live TeamTalk sessions.
+/// Used to ignore benign CoreAudio device-list churn (e.g. Continuity devices)
+/// while still reacting to real route changes (AirPods, unplugged hardware, defaults).
+struct AudioRoutingSnapshot: Equatable {
+    var resolvedInputUID: String?
+    var defaultInputUID: String?
+    var defaultOutputUID: String?
+    var preferredOutputPersistentID: String?
+    var outputPersistentIDInCatalog: Bool
+    var activeInputSampleRate: Double
+}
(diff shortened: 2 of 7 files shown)


math65:
Hey Casey, thanks for the PR! Tested locally and hit two regressions during manual testing that I pushed a follow-up commit (c8770bf) for. Want to make sure my fixes match your intent before merging:

**Regression 1 — changing input/output device in Preferences was a no-op when audio was active.**

`applyAudioPreferences` now calls `extendDeviceChangeSuppressionLocked(duration: 5.0)` right before invoking `restartSoundSystem`. Inside `restartSoundSystem`, the very first thing checked is `if Date() < self.suppressDeviceChangeUntil { return }` — so it bailed every time. The user picks a different mic in Prefs, nothing happens.

Removed the `extendDeviceChangeSuppressionLocked` in `applyAudioPreferences` since `restartSoundSystem` already sets its own 5s window at line 98, after the early-return check.

**Regression 2 — back-to-back device changes within 5s were silently skipped.**

Once #1 was fixed, the next issue: every `restartSoundSystem` call extends suppression to `now + 5s`, so any subsequent explicit call within that window hits the early-return guard. Reproduced by switching input device A → B → A in quick succession; only the first switch took effect.

The early-return on `suppressDeviceChangeUntil` made sense in the pre-PR design where `AppDelegate` called `restartSoundSystem` directly from a debounced hardware-change observer — we needed to coalesce. But now hardware events go through `processAudioHardwareChangeLocked` which has its own suppression check, so the guard inside `restartSoundSystem` is redundant **and** blocks legitimate user-initiated calls.

Removed the early-return; kept the 5s extension inside `restartSoundSystem` (still useful to ignore the CoreAudio notifications generated by recreating the AEC tap).

Let me know if this matches what you had in mind, or if I'm missing context on why the guard was there. Happy to revert and explore alternatives if you prefer.

xogium:
Thanks for testing and for the follow-up in c8770bf — both regressions are real, and your fixes match what I had in mind.

**Regression 1:** Putting `extendDeviceChangeSuppressionLocked` in `applyAudioPreferences` right before `restartSoundSystem` was self-defeating; the early-return in `restartSoundSystem` meant Preferences device changes never actually ran. Dropping the pre-call suppression there is correct — `restartSoundSystem` already extends the window once it starts.

**Regression 2:** The suppression early-return in `restartSoundSystem` made sense when AppDelegate called it on every debounced hardware event, but after routing hardware through `processAudioHardwareChangeLocked` (which has its own suppression check), blocking explicit callers was wrong. User-initiated restarts (Preferences, refresh devices) should always run; only automatic hardware handling should respect the 5s window.

Happy with c8770bf on top of the branch. The separation we wanted is: **explicit paths → always restart; automatic hardware events → snapshot + suppress benign churn**. Your commit preserves that without breaking Prefs.

I’ve ticked the remaining test-plan items on my side where applicable — good to merge from my end when you’re ready.

PR #16, opened 2026-06-16
Add multi-profile support via a New Instance menu

heath-toby:
## Summary

Adds TeamTalk-style profiles to ttaccessible so two instances of the app can run side by side, each with its own saved servers, preferences, passwords, sound packs, and chat history.

- New **New Instance…** item in the app menu (⌘⇧N) shows a picker of existing profiles plus a **New Profile…** button. Selecting a profile launches a separate process via `NSWorkspace.openApplication` with `-profile <slug>` and `createsNewApplicationInstance: true` — no in-app switching, both instances coexist.
- Each profile gets its own `UserDefaults(suiteName:)`, an Application Support subdirectory (`~/Library/Application Support/ttaccessible/profiles/<slug>/`), and a `kSecAttrService` namespace for the keychain.
- `ProfileContext` is resolved lazily from `CommandLine.arguments` on first access, so it's bound before any store is constructed. Stores that already accepted a `UserDefaults` injection now default to `ProfileContext.current.userDefaults`. Stores that hardcoded `.standard` / Application Support paths (`SoundPlayer`, `ChannelChatHistoryStore`, `AnnouncementService`, `FeedbackWindowController`) route through `ProfileContext` too.
- The **Default** profile keeps using `UserDefaults.standard` and the historical Application Support paths — no migration, existing installs see no change.
- Non-default profiles append their display name to the saved-servers and connected-server window titles so multiple running instances are visually distinguishable (e.g. `TeamTalk Servers — Work`).
- New strings localized in en + fr (vouvoiement).

## Architecture

Three new files:

- `Models/ProfileContext.swift` — identifies the active profile and exposes its `UserDefaults`, Application Support root, sound-packs directory, chat-history directory, keychain service name, and a `decorateWindowTitle(_:)` helper. The singleton `current` resolves once from `-profile <slug>` (or the `TTACCESSIBLE_PROFILE` env var) and registers unknown slugs on the fly so a freshly chosen profile works on its very first launch.
- `Services/ProfileRegistry.swift` — stores the list of profiles in a dedicated `UserDefaults(suiteName: "com.math65.ttaccessible.profiles")` shared across instances, so a profile created in one instance shows up in another's picker immediately.
- `AppKit/AppDelegate+NewInstance.swift` — the menu action: shows the picker, prompts for a new name when requested, calls `NSWorkspace.openApplication`. Refuses to launch a second instance of the same profile (which would mean two processes writing to the same suite).

## Test plan

- [x] Builds clean (`Release`, no warnings).
- [x] Manually tested: New Instance picker launches a second process; the second process has its own saved-servers list, preferences, and window title suffix.
- [ ] Default profile data carries over from a pre-PR install (no migration step required, just confirm).
- [ ] Verify keychain prompts on first server-password access in a new profile.
- [ ] Verify French strings render with `defaults write -g AppleLanguages -array fr en` then relaunch.

### DIFF
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -107,9 +107,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var pushToTalkModeCancellable: AnyCancellable?
 
     func applicationDidFinishLaunching(_ notification: Notification) {
+        ProfileInstanceLock.acquire(for: ProfileContext.current)
         AudioLogger.clear()
         let sdkVersion = String(cString: TT_GetVersion())
-        AudioLogger.log("App launched — TeamTalk SDK %@", sdkVersion)
+        AudioLogger.log("App launched — TeamTalk SDK %@ — profile %@", sdkVersion, ProfileContext.current.slug)
         #if DEBUG
         _ = AudioPCMResamplerSelfTest.runAll()
         #endif
@@ -318,6 +319,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     func applicationWillTerminate(_ notification: Notification) {
+        ProfileInstanceLock.release(for: ProfileContext.current)
         connectionController.disconnectSynchronously()
         // The TeamTalk SDK's internal reactor thread sometimes outlives
         // TT_CloseTeamTalk and crashes when exit()'s static destructors race
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -380,7 +380,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
            window.contentViewController is SavedServersViewController == false {
             let viewController = makeSavedServersViewController()
             window.contentViewController = viewController
-            window.title = L10n.text("savedServers.window.title")
+            window.title = ProfileContext.current.decorateWindowTitle(L10n.text("savedServers.window.title"))
         }
 
         menuState.setMode(.savedServers)
@@ -435,7 +435,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         }
 
         savedServersWindowController?.window?.contentViewController = viewController
-        savedServersWindowController?.window?.title = L10n.format("connectedServer.window.title", session.displayName)
+        savedServersWindowController?.window?.title = ProfileContext.current.decorateWindowTitle(
+            L10n.format("connectedServer.window.title", session.displayName)
+        )
         menuState.setMode(.connectedServer)
         menuState.setHasSelection(false)
         if shouldActivateWindow {
--- a/App/ttaccessible/AppKit/AppDelegate+NewInstance.swift
+++ b/App/ttaccessible/AppKit/AppDelegate+NewInstance.swift
@@ -1,0 +1,137 @@
+//
+//  AppDelegate+NewInstance.swift
+//  ttaccessible
+//
+
+import AppKit
+
+@MainActor
+extension AppDelegate {
+    /// Show the picker that lets the user launch a separate process bound to
+    /// another profile. Triggered from the "New Instance…" menu item.
+    func openNewInstanceDialog() {
+        let registry = ProfileRegistry.shared
+        let existing = registry.listAll()
+        let currentSlug = ProfileContext.current.slug
+
+        let alert = NSAlert()
+        alert.alertStyle = .informational
+        alert.messageText = L10n.text("profile.newInstance.title")
+        alert.informativeText = L10n.text("profile.newInstance.message")
+        alert.addButton(withTitle: L10n.text("profile.newInstance.launch"))
+        alert.addButton(withTitle: L10n.text("profile.newInstance.createNew"))
+        alert.addButton(withTitle: L10n.text("common.cancel"))
+
+        let popup = NSPopUpButton(frame: NSRect(x: 0, y: 0, width: 320, height: 26), pullsDown: false)
+        popup.setAccessibilityLabel(L10n.text("profile.newInstance.picker.accessibilityLabel"))
+        for entry in existing {
+            let title: String
+            if entry.slug == currentSlug {
+                title = L10n.format("profile.newInstance.picker.currentSuffix", entry.displayName)
+            } else {
+                title = entry.displayName
+            }
+            popup.addItem(withTitle: title)
+            popup.lastItem?.representedObject = entry.slug
+        }
+        if popup.numberOfItems == 0 {
+            // Should never happen — the registry always includes the default entry.
+            popup.addItem(withTitle: ProfileContext.defaultDisplayName)
+            popup.lastItem?.representedObject = ProfileContext.defaultSlug
+        }
+
+        // Default the picker to the first non-current profile if any exist.
+        if let firstOther = existing.firstIndex(where: { $0.slug != currentSlug }) {
+            popup.selectItem(at: firstOther)
+        }
+
+        alert.accessoryView = popup
+
+        let response = alert.runModal()
+        switch response {
+        case .alertFirstButtonReturn:
+            guard let slug = popup.selectedItem?.representedObject as? String else { return }
+            launchInstance(forSlug: slug)
+        case .alertSecondButtonReturn:
+            promptCreateAndLaunchProfile()
+        default:
+            return
+        }
+    }
+
+    private func promptCreateAndLaunchProfile() {
+        let alert = NSAlert()
+        alert.alertStyle = .informational
+        alert.messageText = L10n.text("profile.create.title")
+        alert.informativeText = L10n.text("profile.create.message")
+        alert.addButton(withTitle: L10n.text("profile.create.launch"))
+        alert.addButton(withTitle: L10n.text("common.cancel"))
+
+        let field = NSTextField(frame: NSRect(x: 0, y: 0, width: 320, height: 24))
+        field.placeholderString = L10n.text("profile.create.placeholder")
+        field.setAccessibilityLabel(L10n.text("profile.create.field.accessibilityLabel"))
+        alert.accessoryView = field
+        alert.window.initialFirstResponder = field
+
+        guard alert.runModal() == .alertFirstButtonReturn else { return }
+
+        let rawName = field.stringValue.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard rawName.isEmpty == false else {
+            presentNewInstanceError(message: L10n.text("profile.create.error.empty"))
+            return
+        }
+
+        let slug = ProfileContext.normalizeSlug(rawName)
+        guard slug.isEmpty == false else {
+            presentNewInstanceError(message: L10n.text("profile.create.error.invalid"))
+            return
+        }
+        guard slug != ProfileContext.defaultSlug else {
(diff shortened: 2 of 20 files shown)


math65 (changes requested):
Thanks for this — really clean work. I pulled the branch, it builds clean in Release (the three new files are picked up automatically by the synchronized group, no `pbxproj` changes needed), and the store migration is complete: every `UserDefaults.standard` usage and every Application Support path is now routed through `ProfileContext`, with the Default profile preserving the historical behavior so existing installs need no migration. Localization is in place (French uses vouvoiement). The overall design — one process per profile via `-profile <slug>`, isolation through a dedicated `UserDefaults` suite + App Support subdirectory + keychain service namespace — is the right approach.

A few things I'd like addressed or at least discussed before merging.

### Please add: profile deletion (and ideally rename)

`ProfileRegistry` is currently create-only — there's no way to delete or rename a profile, and no UI for it. Once a profile is created it stays in the picker forever, and a typo'd name can't be fixed. For a general-purpose feature this is going to bite users quickly (everyone creates a "test" profile at some point). Deletion is the important one; rename is nice-to-have.

Note that a proper delete needs to clean up everything the profile owns, not just the registry entry:
- remove the registry entry,
- delete the `profiles/<slug>/` Application Support subdirectory,
- remove the per-profile `UserDefaults` suite (`removePersistentDomain(forName:)`),
- purge the keychain items under that profile's `kSecAttrService` namespace.

And it should refuse to delete the currently-running profile / the Default profile.

### Worth a look

1. **Cross-instance double-launch of the same profile.** The guard in `launchInstance(forSlug:)` only blocks the *current* process's profile. From the "Work" instance you can still pick "Default" and launch a second Default process while one is already running — exactly the "two processes writing the same suite" case the comment warns against. There's no cross-process check for already-running profiles (I realize that's hard to do robustly since launch arguments aren't readily readable across processes). At minimum worth a comment documenting the limitation; ideally a best-effort check.

2. **Sparkle relaunch drops `-profile`.** After an update, Sparkle relaunches the app without `-profile <slug>`, so a non-default instance comes back as Default. Two instances also both run Sparkle update checks/installs in parallel, which bumps into Sparkle's single-instance assumptions. Edge case, but worth noting.

3. **Shared sandbox container.** Both processes share `~/Library/Containers/com.math65.ttaccessible/`. The `UserDefaults` suites are correctly isolated (separate plists), but `audio.log` is shared — each instance wipes it on launch, so logs from two instances overwrite/interleave each other. Minor, but it'll make audio debugging confusing once two instances are involved.

### Nits

- `ProfileContext.current` is a `var` that's never reassigned — could be `let`.
- The doc comment on `applicationSupportRoot` mentions `~/Library/Application Support`, but under the sandbox this resolves to the container path. Functionally fine, just slightly misleading wording.
- Possible read-modify-write race on the shared registry suite if two instances register simultaneously (low probability).

### Not a code issue, but the thing I most want to validate

Two instances each run the full TeamTalk SDK + mic capture + the AEC speaker tap (a system-wide process tap + private aggregate device). Running two side-by-side means mic contention, duplicate aggregate devices, and `restartSoundSystem()` firing in both on any device change — the audio pipeline was never designed for two simultaneous instances. This is the core use case of the feature, so I want to test it live before merging. No action needed from you here; just flagging why I'm not merging immediately.

---

Anything else on your end — design decisions you're unsure about, things you ran into, or follow-ups you were planning? Happy to discuss any of the above before you start changing things.

heath-toby:
Thanks for the thorough review. Follow-up pushed as db6e789. Summary of what's addressed and what isn't:

### Done — profile deletion + rename

- New **Manage Profiles…** item under the same App menu group as New Instance (no shortcut). Picker of custom profiles + Rename / Delete buttons. Empty registry shows a friendly placeholder dialog instead of opening an empty picker.
- **Rename** updates the display name only. The slug is immutable because it backs file paths, the UserDefaults suite name, and the keychain service — changing it would require migrating all of that, which I'd rather not do casually.
- **Delete** runs `ProfileContext.purgeStorage(forSlug:)` which:
  - removes the `profiles/<slug>/` Application Support subdirectory,
  - calls `removePersistentDomain(forName:)` on the suite (plus a belt-and-braces second call on a throwaway `UserDefaults()` instance — earlier macOS occasionally left the backing plist behind otherwise),
  - calls `SecItemDelete` with `kSecClass: kSecClassGenericPassword, kSecAttrService: <profile-service>` to purge keychain items,
  - then drops the registry entry.
- Refuses to delete the **currently-running** profile (`ProfileContext.current`) and refuses to delete any profile **another live instance** holds (checked via the new lock — see below). Confirmation alert has Cancel as the default Return action so a stray Enter can't blow away a profile.

### Done — cross-instance double-launch

`ProfileInstanceLock` writes the running PID to `Application Support/ttaccessible/instance-locks/<slug>.pid` on launch and removes it on graceful exit. `launchInstance(forSlug:)` reads the file and uses `kill(pid, 0)` as a liveness probe — stale locks from crashes auto-clear on the next check. From the Work instance, picking Default now gets refused if a Default process is already running, and vice versa.

Limitations are documented in the file header:
- crash leaves a stale file, but the liveness probe handles that on the next attempt;
- PID recycling could in theory false-positive (worst case: user sees a spurious refusal and deletes the lock file);
- two simultaneous acquire calls race, but per-profile isolation already tolerates last-writer-wins on UserDefaults.

### Done — shared audio.log

`AudioLogger` now routes non-default profiles to `audio-<slug>.log` in the same `~/Library/Logs/TTAccessible/` directory. Default keeps `audio.log` so the audio-log skill (and the feedback bug-report attachment) keep working unchanged for the common case. Launch log line also records the bound profile slug, which should help if a future bug report comes from a non-default instance.

### Done — nits

- `ProfileContext.current` is now `let`.
- `applicationSupportRoot` doc clarifies that under the App Sandbox it resolves inside the container, not literal `~/Library/Application Support/` — functionally fine since every instance of the bundle sees the same container.
- `ProfileRegistry` mutations hold an in-process `NSLock` and call `synchronize()` after write. Cross-process simultaneous registrations still have a small read-modify-write window (inherent to UserDefaults — they don't expose CAS) but registration is rare and user-initiated, so I left it at "documented" rather than introducing a file-based mutex.

### Not addressed — Sparkle relaunch drops -profile

Genuinely tricky. Sparkle relaunches the bundle without preserving the launch args, and there's no `pathToRelaunchForUpdater(_:)`-style hook for additional arguments that I could find. The options I considered:

1. **Persist the active slug to a known file** the next-launched process can read — falls apart with two non-default instances running, since whichever wrote last wins.
2. **Disable Sparkle on non-default instances** — clean for the dual-update case but breaks the "user only ever runs Work, never Default" workflow.
3. **Write a slug marker keyed by PID** — Sparkle's relaunch doesn't preserve PID either.

I'd rather ship the rest behind it than land a half-fix; happy to break this out into a follow-up issue if you agree, with the dual-Sparkle update prompts also called out there.

### About the audio side

Totally agree on holding the merge until you can sanity-check two SDK instances together. Nothing I can do for that one from this side — it's exactly the kind of thing that needs real hardware and a real second user to talk to. Let me know what you find.

math65 (approved):
The follow-up (db6e789) addresses everything I raised, and cleanly:

- **Delete + rename** via "Manage Profiles…" — `purgeStorage` tears down the App Support subdir, the `UserDefaults` suite, and the keychain items, and the registry entry is dropped after. I confirmed the keychain purge is correct: `ServerPasswordStore` doesn't use the data-protection keychain, and neither does the purge query, so they hit the same domain and the `SecItemDelete` actually removes the items. Refusing to delete the current/another-live profile + Cancel as the default Return action are nice touches.
- **`ProfileInstanceLock`** (PID file + `kill(pid,0)` liveness, self-guard, stale cleanup) — solid, and the limitations are honestly documented.
- **Per-profile `audio-<slug>.log`**, Default keeps `audio.log` so the audio-log tooling and bug-report attachment stay intact for the common case.
- All nits resolved.

Sparkle dropping `-profile` on relaunch: agreed it's genuinely awkward and not worth a half-fix — let's track it as a follow-up issue (along with the dual-instance Sparkle update prompts).

I tested two instances side by side and the audio held up fine on my end, which was my one blocking concern. Merging now. Thanks for the thorough, well-documented work!

math65:
Quick follow-up on the one piece we deferred — the Sparkle relaunch dropping `-profile`. It turned out to be solvable cleanly, so I went ahead and fixed it on `main` (ef1535a), tracked in #17.

The trick was to treat it as a **one-shot handoff token** rather than a persisted "active profile" (which is what made your option 1 ambiguous with two non-default instances): `updaterWillRelaunchApplication(_:)` writes the current slug to a short-lived file in the shared coordination dir, and the next launch consumes it — deleted on read, ignored after 60s. Since it's written by the exact instance Sparkle is relaunching and read exactly once, parallel non-default instances don't collide; they keep their in-memory `-profile` and aren't relaunched. Tested both ways (fresh token rebinds, stale token falls back to Default).

The only bit still open in #17 is the dual-instance Sparkle update-prompt question, which is minor. Thanks again for the great feature and the thorough write-ups — they made this easy to reason about.

math65:
Heads-up: after merging, I noticed the profile isolation wasn't actually holding — a freshly created profile showed the Default profile's saved servers. Digging in, there were two separate bugs underneath, both producing the same misleading "looks fine" symptom (a second window with a server list), which is exactly why it slipped past your manual test and my review. Fixed on `main` in 4a1d42a.

**1. `UserDefaults(suiteName:)` falls through to the app's bundle domain.**
In-process, a suite's search list still includes the main bundle's persistent domain. With the Default profile living in `.standard`, any key a new profile hadn't written yet was read back from the Default's data — so a brand-new profile transparently showed the Default's servers/preferences (passwords excepted, since the keychain namespace *is* per-profile, which is what made it inconsistent).

Fix: the Default profile now gets its own suite too, and a one-time, flag-guarded migration moves the Default's app keys out of the bundle domain (leaving Sparkle/AppKit framework keys in place). Other profiles' fall-through then lands on nothing → real isolation, no data loss for existing installs.

**2. `NSWorkspace.openApplication` drops `arguments` *and* `environment` for a sandboxed self-launch.**
This was the bigger one. The child instance never received `-profile <slug>` (nor the `TTACCESSIBLE_PROFILE` env fallback) — both are silently dropped when a sandboxed app launches another instance of its own bundle. So every "new instance" came up bound to the Default profile. I confirmed it with logging: parent requests `args=-profile <slug>`, child launches with `args=<empty>` and resolves `default`.

Fix: the target profile (plus a "start disconnected" flag) is handed off via a one-shot file token the child consumes at launch — the same mechanism as the Sparkle relaunch handoff, now generalized. Arguments/environment are kept as belt-and-braces for non-sandbox/external launches where they *do* come through.

While there, I also made it work like the Qt client: launching the **current** profile is now allowed (a clone that shares its config and starts disconnected, for connecting to a second server), and `ProfileInstanceLock` is kept only to guard deletion rather than to refuse a launch.

None of this is a knock on the work — the design and architecture were right, these are two genuinely sticky platform gotchas that happen to fail in a way that imitates success. Thanks again; the feature's in great shape now. The Sparkle-relaunch follow-up (#17) is still the only open item.

math65:
Small correction to my note above: only the *minor* part of #17 was still open. The Sparkle-relaunch-loses-profile concern itself is fixed by the same handoff token (ef1535a / 4a1d42a). I've closed #17 — the leftover (duplicate update prompt when two instances run) is accepted as-is, matching the Qt client.

PR #19, opened 2026-06-18
Add "No output device" option to audio preferences

heath-toby:
## Summary

Adds a third item to the Output device picker in the Audio preferences pane: **No output device**. Picking it makes the profile transmit and stay connected without ever opening a sound output device — useful for streaming profiles, server-monitor windows, or any setup where another logged-in profile is already carrying the audio.

## Behaviour

- Selecting **No output device** persists `AudioDevicePreference.noOutput` (sentinel `persistentID = "__no_output__"`).
- `ensureDirectOutputAudioReadyLocked` is the single entry point every code path uses to open the SDK output. It now early-returns (logs + leaves `outputAudioReady = false`) when the preference is no-output. Channel-join, reconnect, restart-sound-system, mic-start, and AEC paths all flow through this function, so they inherit the behaviour without further changes.
- `restartSoundSystem`'s output re-open guard is loosened: it used to gate on "was output open before the restart?", which would leave the device closed when switching FROM no-output TO a real device while the mic was active. New gate: "was open before, OR is the current preference a real device?" The short-circuit in `ensureDirectOutputAudioReadyLocked` keeps the unconditional case safe.

## What this is not

- Not a runtime mute — master mute already exists for that.
- Not a way to disable input. The input picker is unchanged; no-output applies to output only.
- Not Codable-breaking. Existing saved preferences with `persistentID == nil` still mean system default.

## Test plan

- [x] Builds clean in Release, no new warnings.
- [ ] Pick "No output device" while disconnected → connect to a server → channel join: no playback, mic transmits, you appear in the channel.
- [ ] Pick "No output device" while connected and in a channel → audio playback should stop without disconnecting.
- [ ] Switch from "No output device" back to a real device while in a channel with mic active → playback should resume after the SDK restart.
- [ ] Restart the app on a profile saved with no-output → connection works, no SDK output init attempt in the audio log (look for the `no-output preference — skipping init` line).
- [ ] French UI: picker shows "Aucune sortie audio".

### DIFF
--- a/App/ttaccessible/Models/AudioDevicePreference.swift
+++ b/App/ttaccessible/Models/AudioDevicePreference.swift
@@ -8,12 +8,24 @@
 import Foundation
 
 struct AudioDevicePreference: Codable, Equatable {
+    /// Sentinel `persistentID` that means "do not initialize an output device
+    /// at all" (rather than picking a real device or the system default).
+    /// Used by the Audio preferences pane's output picker so streaming
+    /// profiles can connect, transmit, and hear nothing — the main profile
+    /// carries the audio.
+    static let noOutputSentinelID = "__no_output__"
+
     var persistentID: String?
     var displayName: String?
 
     static let systemDefault = AudioDevicePreference(persistentID: nil, displayName: nil)
+    static let noOutput = AudioDevicePreference(persistentID: noOutputSentinelID, displayName: nil)
 
     nonisolated var usesSystemDefault: Bool {
         persistentID?.isEmpty != false
+    }
+
+    nonisolated var usesNoOutput: Bool {
+        persistentID == AudioDevicePreference.noOutputSentinelID
     }
 }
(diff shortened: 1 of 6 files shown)


math65:
Merged and shipped in **v1.6.0** 🎉 — thank you for this!

I tested it live with two instances on the same channel: the no-output profile stays connected and its microphone transmits fine while never opening a sound output, exactly as intended. I added two tiny cosmetic follow-ups on top (789bbf3): a dedicated "Output disabled" audio status instead of "Audio not initialized" (so VoiceOver doesn't read a deliberate choice as an error), and the feedback diagnostic now shows "aucune sortie audio" instead of the raw `__no_output__` sentinel.

Really appreciate the clean, well-documented PR — the single-entry-point design through `ensureDirectOutputAudioReadyLocked` made it easy to review. 🙏

PR #20, opened 2026-06-21
Audio reliability + accessibility overhaul (device-switch deadlock cure, channel mixer, faster connect, VoiceOver fixes)

rfiorentino1:
Hi — first off, thanks for ttaccessible; it's a genuinely useful app and a
great base to build on.

This is a large PR that bundles a chunk of work I did on a fork: audio-engine
reliability, a per-user mixer, faster connects, and a broad VoiceOver/accessibility
pass. It's branched on top of current `main` (v1.7.0-beta.1 — I merged your BearWare
web-login work in, and it's included/working here).

I know this is a lot to land in one shot. **I'm very happy to split it into smaller,
themed PRs** (accessibility-only, audio-only, etc.) if that's easier to review — just
say the word and I'll carve it up.

## Highlights

**Audio reliability**
- **Cure for the output-device-switch deadlock.** `TT_CloseSoundOutputDevice`
  intermittently deadlocks on macOS when switching the active output device. This
  routes the SDK's output to the TeamTalk virtual device and renders incoming audio
  ourselves through a per-user mixing engine bound to a CoreAudio device, so device
  switches never close a live SDK output device. Same approach removes the analogous
  input-side hazard.
- **Stable audio-device memory.** Device preferences were keyed by the SDK's integer
  device index (`nDeviceID`), which reshuffles across launches/hot-plug, so the app
  could silently select the wrong device on restart. Now keyed by the stable CoreAudio
  device UID, with a one-time migration of existing prefs.
- Lower receive-path latency and a perceptually-even volume curve.

**Channel mixer**
- A live per-user desk: per-person voice/media volume, continuous pan, mute/solo, with
  full keyboard control — built for VoiceOver users.

**Faster connect**
- The SDK's PortAudio backend runs a slow device probe at init (~13s on large rigs).
  A targeted patch + instance prewarm/reuse cuts first-connect dramatically.

**Accessibility / VoiceOver**
- De-duplicated the Preferences sidebar reading, labeled each preferences scroll area,
  Escape closes the Preferences window, "Hear myself" reflects on-state as selected,
  collapsed several redundant double-readouts, and more.
- Fixed a small gap in your new BearWare pane (it wasn't passing the accessibility
  label the labeled-scroll-area change now requires).

## Things worth flagging honestly
- **This is an architectural change**, not just patches — the SDK no longer owns the
  physical audio devices in the bypass path. It's been stable in daily use on my end,
  but it deserves real review and testing, not a rubber stamp.
- **It patches the bundled TeamTalk SDK dylib** (`scripts/patch-sdk-portaudio.py`, run
  from the SDK download script) to fix the slow startup probe. Worth a look if you'd
  rather handle that differently.
- **It renames the app to "ttAccessible"** (`CFBundleDisplayName` + a few strings).
  This is an accessibility fix: VoiceOver reads "ttaccessible" as one run ("taccessible")
  but reads "ttAccessible" correctly as "TT Accessible." Easy to drop if you'd prefer to
  keep the name as-is.

Happy to adjust, split, or explain any of it. Thanks for taking a look!


math65:
Thanks so much for this, Rocco — and for the heads-up too. I went through the whole PR and built/tested it locally, and it's excellent work. The engineering judgment behind it (the device-switch deadlock cure, the PortAudio probe patch) is clearly the real deal.

Quick context on me: I'm not a professional dev — I build this while learning, with a lot of AI assistance, and I really want to understand everything that goes in. So I'd love to take you up on your offer to split this into smaller, themed PRs. It'll let me review and test each piece properly instead of merging ~3,400 lines I only half understand. Here's the grouping I'd suggest — but you know the code better than I do, so push back freely:

1. **Accessibility / VoiceOver + Preferences cleanup + localization** — low-risk, high-value. I'd merge this first and fast.
2. **App rename to "ttAccessible"** — yes, I'm happy to take this one; the VoiceOver pronunciation reasoning convinced me. Since it's an accessibility fix anyway, it can ride along with group 1, or stay a tiny standalone PR — your call.
3. **The audio work as one cohesive set**: output rearchitecture (virtual-device bypass + per-user render engine + UID-keyed device prefs), the channel mixer, the advanced-mic preview routing, and the faster-connect (SDK dylib patch + prewarm).

On grouping #3: I think these need to travel together rather than as fully independent PRs, because they're coupled — the mixer's pan/solo and the mic preview both depend on the new `OutputAudioRenderEngine`, and the SDK probe-skip patch is only fully safe once output goes through the virtual device (on current `main`, output still opens a real PortAudio device). If you see clean seams inside that set, stacked PRs would be perfect — but I wouldn't want to merge the patch or the mixer without the rearchitecture underneath.

That audio set is also the part I want to test most before committing, since it fundamentally changes how the audio flows. One specific thing I'd love your take on: in the new path, per-user volume and mute still go through the SDK (`TT_SetUserVolume` / `TT_SetUserMute`) while the decoded per-user PCM is what we render — do those still affect what the user actually hears? That's my main open question on the architecture.

Two small notes:
- I loved the dylib patch. I added a small safety check to the script so a future SDK layout change fails loudly instead of risking a corrupted binary.
- This branch is based on 1.7.0-beta.1; `main` has since moved to beta.2 (a launch-speed + nickname fix). I merged current `main` in locally and it was clean and built fine, so when you re-cut the smaller PRs you may want to rebase on latest `main`.

Thank you again, genuinely — really happy to collaborate on this going forward.

rfiorentino1:
Splitting this up as discussed — replaced by two focused PRs:
- #21 — accessibility, VoiceOver & Preferences cleanup (+ rename to tt-Accessible)
- #22 — the audio set: output-bypass engine, channel mixer, faster connect

Closing this in favor of those. Thanks!

PR #21, opened 2026-06-23
Accessibility & VoiceOver: Preferences cleanup, VO-Space activation, and rename to tt-Accessible

rfiorentino1:
Hi Mathieu — this is the first of the smaller, themed PRs we discussed splitting out of #20: the low-risk, high-value accessibility group, rebased on current `main` (v1.7.0-beta.5). No audio or architecture changes here — those come in a later PR.

## Included

- **Preferences VoiceOver cleanup** — de-dupe the sidebar, Escape closes the window, label each scroll area, remove redundant static text, and collapse the sound-event switch labels into one element.
- **"Hear myself" toolbar button** reflects its on-state as VoiceOver-selected.
- **VO-Space (VoiceOver press) joins servers and channels** — previously only Return worked, and only while interacting with the table. Now the accessibility press action on the table/outline and its cells triggers the same connect / join.
- **Volume dialog** — stop VoiceOver reading the redundant "%" readout next to each slider (kept visible for sighted users; the slider already announces the value).
- **Collapse display name** when nickname == username (no double readout).
- **App rename to "tt-Accessible"** — note the hyphen, not "ttAccessible" as I first suggested. Eloquence and some other synthesizers mis-read the camelCase as a single run; a hyphen reads correctly across voices. `CFBundleName` / `CFBundleDisplayName` + user-facing copy only — the bundle identifier (`com.math65.ttaccessible`) and on-disk product name are unchanged.
- **BearWare pane fix** — `PreferencesBearWareView` now passes the scroll-area accessibility label that the labeled-scroll-area change makes required (the pane was added upstream after that change, so it was missing it).

Built against current `main`. Happy to adjust, squash further, or split.

### DIFF
--- a/App/ttaccessible-Info.plist
+++ b/App/ttaccessible-Info.plist
@@ -62,7 +62,9 @@
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundleName</key>
-	<string>$(PRODUCT_NAME)</string>
+	<string>tt-Accessible</string>
+	<key>CFBundleDisplayName</key>
+	<string>tt-Accessible</string>
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
@@ -72,13 +74,13 @@
 	<key>NSHumanReadableCopyright</key>
 	<string></string>
 	<key>NSAppleEventsUsageDescription</key>
-	<string>TTAccessible can optionally control VoiceOver to read incoming messages while the app is in the background.</string>
+	<string>tt-Accessible can optionally control VoiceOver to read incoming messages while the app is in the background.</string>
 	<key>NSMicrophoneUsageDescription</key>
-	<string>TTAccessible uses the microphone to talk in TeamTalk channels.</string>
+	<string>tt-Accessible uses the microphone to talk in TeamTalk channels.</string>
 	<key>NSAudioCaptureUsageDescription</key>
-	<string>TTAccessible can capture audio from other applications to stream it to TeamTalk channels.</string>
+	<string>tt-Accessible can capture audio from other applications to stream it to TeamTalk channels.</string>
 	<key>NSScreenCaptureUsageDescription</key>
-	<string>TTAccessible can capture audio from specific applications to stream it to TeamTalk channels.</string>
+	<string>tt-Accessible can capture audio from specific applications to stream it to TeamTalk channels.</string>
 	<key>SUFeedURL</key>
 	<string>https://math65.github.io/ttaccessible/appcast.xml</string>
 	<key>SUPublicEDKey</key>
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -1355,6 +1355,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     func toggleHearMyself() {
         guard menuState.mode == .connectedServer else { return }
         connectionController.toggleHearMyself { [weak self] enabled in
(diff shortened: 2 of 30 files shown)


math65 (commented):
Thanks Rocco — this is exactly the clean, low-risk split we wanted. I ran a deep (multi-agent) review plus a local build + the unit-test suite (all green), and traced the riskier paths myself.

**No blockers.** Verified safe:
- **VO-Space activation** — no stale-node risk. The full `reloadData()` re-binds every cell's `onPress` to a fresh node; the partial `reloadData(forRowIndexes:)` (audio-runtime updates) only touches `.user` rows, where `id`/`displayName` are stable, so VO-Space can't act on the wrong user or channel. Delegate methods wired, `ServerTreeNode` switch exhaustive, `appDelegate` is the injected reference (not the nil-prone `NSApp.delegate` cast).
- **SwiftUI a11y refactor** — every relabeled `Toggle` has its compensating `.accessibilityLabel`, no `Picker` left unlabeled, all 10 `PreferencesPaneScrollView` call sites pass the new required parameter.
- **Rename** — confirmed `ttaccessible.app` (on disk), the bundle id, the sandbox container, the log path, `build.sh`, Sparkle and notarization are all decoupled from `CFBundleName`, so nothing structural moves.
- **`PercentValueView`** — nice solution to the NSAlert-reads-the-accessory problem.

**On the client name:** I checked that `clientName` (= `CFBundleName`) is sent to the server and shown to other users, so they'll now see "tt-Accessible" instead of "TTAccessible". That's **intentional on my side — please keep it as is**, no need to decouple it.

**Nits — already handled, nothing to do on your end:** I pushed a small polish commit straight to this branch (`bc2278e`): README copy → "tt-Accessible", de-duped the `PercentValueView` doc-comment, fixed a stale comment in `+Identity.swift`, and corrected an accessibility-label key in `PreferencesNotificationsView` (dead code, but tidy).

**One question (the only real point):** the cell-level press (`PressActionTextField.onPress`) acts on the explicit pressed node, but the table/outline-level `accessibilityPerformPress()` overrides route back through the **selection**-based path (`performDefaultAction()` / `connectSelectedServer()`), which no-ops when nothing is selected — exactly the "no selection without interaction" case your own comments say you're solving. In practice it's masked because `selectDefaultRow()` keeps a row selected, so I don't think it's a real bug. Is the selection-based table-level press deliberate, or would you rather have it honor the same "act on the focused/pressed row regardless of selection" contract as the cells? Happy either way — just flagging the inconsistency.

Holding the merge until you've weighed in on that one. The closely-related nit (the outline cell's VoiceOver custom actions are still selection-based) is part of the same question, so I left it untouched.

rfiorentino1:
Thanks Mathieu — and thanks for the deep review and the polish commit.

On the one open question: the table/outline-level `accessibilityPerformPress()` going through the selection is **deliberate**. It's meant to mirror the Return-key path exactly — both route through `performDefaultAction()` / `connectSelectedServer()` on the selected row — so VO-Space *at the container level* behaves identically to pressing Enter, which is the contract I wanted there. And because `selectDefaultRow()` always keeps a row selected, the "no selection" no-op never actually occurs in practice: in testing VO-Space activates correctly whether VoiceOver is focused on the table/outline itself or interacting down on a cell.

The cell-level press acts on the explicit pressed node because at that level the target is unambiguous. So the two paths read their target from different places but converge on the same behavior, and I'm happy to leave both as-is. Good to merge from my side. 🙏

math65:
Merged. 🎉 Thanks Rocco — clean split, well-reasoned design choices. This ships in the next beta (1.7.0-beta.6) so AppleVis testers get the VoiceOver improvements. #22 (audio) is next: I'll rebase it on main, then do the deep review + my own listening tests before it goes to a separate beta. Appreciate the collaboration.

PR #22, opened 2026-06-23
Audio: output-bypass engine, channel mixer, faster connect (device-switch deadlock cure)

rfiorentino1:
Hi Mathieu — the audio set we discussed: the cohesive unit you said you'd want to test with a couple of people. **Stacked on #21** (the accessibility PR) — please merge #21 first; until then this PR's diff also shows #21's commits, and it'll collapse to audio-only once #21 lands.

## What it does

- **Output-bypass render engine** — the SDK points at the TeamTalk *virtual* output device, and we render each remote user's decoded PCM ourselves via `OutputAudioRenderEngine`. Switching the active output device never closes a live SDK output device → cures the intermittent `TT_CloseSoundOutputDevice` deadlock on macOS. The same approach removes the input-side hazard.
- **CoreAudio UID-keyed device prefs** — the device catalog is built from CoreAudio and remembered by stable UID across launches/hot-plug (not the SDK's reshuffling integer index, which could silently select the wrong device on restart).
- **Channel Mixer** — a live per-user voice/media/pan/mute/solo desk embedded in the main window, with full keyboard control, built for VoiceOver.
- **Faster connect** — a targeted patch to PortAudio's slow startup device probe (`scripts/patch-sdk-portaudio.py`, run from `download-sdk.sh`) plus instance prewarm/reuse: ~13s → ~0.2s first connect on large rigs.

## Your earlier question, answered

Per-user voice **and** media-file volume and mute all remain audible through `TT_SetUserVolume` / `TT_SetUserMute` in the bypass path (confirmed by listening) — the decoded per-user PCM we render reflects those SDK settings.

## Mixer keyboard shortcuts

While VoiceOver is focused on a mixer strip:

- **↑ / ↓** — focused user's **voice** volume
- **⌘↑ / ⌘↓** — focused user's **media** volume (or **master output** volume when the cursor is *not* on a strip)
- **← / →** — focused user's **pan**
- **v** — announce volume (tap) · reset to 50% (double-tap)
- **p** — announce pan (tap) · reset to center (double-tap)
- **m** — announce mute (tap) · toggle mute (double-tap)
- **s** — announce solo (tap) · toggle solo (double-tap)

Heads-up: I forgot to add any in-app documentation for these — no help tags, no shortcuts reference in the UI — so right now they're only written down here. Making them discoverable in-app (help tags / a shortcuts list, ideally VoiceOver-friendly) is a good follow-up; happy to do that in a separate PR.

## Worth flagging honestly

- This is an **architectural change**, not just patches — it's been stable in daily use here, but it deserves real multi-person testing (exactly what you wanted to do), not a rubber stamp.
- It **patches the bundled TeamTalk SDK dylib** — you already added a safety check to the script so a future SDK layout change fails loudly; thank you for that.
- One open item we're still chasing: on a heavy aggregate / Rogue Amoeba Loopback setup, a CoreAudio device-list storm can disrupt other virtual audio devices. It's under investigation; the app's own render engine stays healthy through it.

Happy to split this further or adjust anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,13 @@ DerivedData/
 *.mode2v3
 *.perspectivev3
 xcuserdata/
+# Auto-generated shared scheme (not tracked upstream)
+App/ttaccessible.xcodeproj/xcshareddata/xcschemes/
 
 # Build artifacts
 BuildArtifacts/
+build_verify/
+*.slow-v5.22a-backup
 
 # macOS
 .DS_Store
--- a/App/ttaccessible-Info.plist
+++ b/App/ttaccessible-Info.plist
@@ -62,7 +62,9 @@
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundleName</key>
-	<string>$(PRODUCT_NAME)</string>
+	<string>tt-Accessible</string>
+	<key>CFBundleDisplayName</key>
+	<string>tt-Accessible</string>
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
@@ -72,13 +74,13 @@
 	<key>NSHumanReadableCopyright</key>
 	<string></string>
 	<key>NSAppleEventsUsageDescription</key>
-	<string>TTAccessible can optionally control VoiceOver to read incoming messages while the app is in the background.</string>
+	<string>tt-Accessible can optionally control VoiceOver to read incoming messages while the app is in the background.</string>
 	<key>NSMicrophoneUsageDescription</key>
-	<string>TTAccessible uses the microphone to talk in TeamTalk channels.</string>
+	<string>tt-Accessible uses the microphone to talk in TeamTalk channels.</string>
 	<key>NSAudioCaptureUsageDescription</key>
-	<string>TTAccessible can capture audio from other applications to stream it to TeamTalk channels.</string>
+	<string>tt-Accessible can capture audio from other applications to stream it to TeamTalk channels.</string>
 	<key>NSScreenCaptureUsageDescription</key>
-	<string>TTAccessible can capture audio from specific applications to stream it to TeamTalk channels.</string>
+	<string>tt-Accessible can capture audio from specific applications to stream it to TeamTalk channels.</string>
 	<key>SUFeedURL</key>
 	<string>https://math65.github.io/ttaccessible/appcast.xml</string>
 	<key>SUPublicEDKey</key>
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -143,6 +143,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             self?.preloadPreferencesWindow()
         }
         hasFinishedLaunching = true
+        // Create the first TeamTalk instance in the background now. TT_InitTeamTalkPoll
+        // enumerates all CoreAudio devices (~12 s on a large rig); prewarming it here
+        // keeps that cost off the connect path so connecting is fast.
+        connectionController.prewarmConnection()
         handleLaunchTTFilesIfNeeded()
         processPendingTTFileURLsIfPossible()
         syncSparkleAutoCheckPreference()
@@ -1514,6 +1518,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         }
         feedbackWindowController?.show()
     }
+
 
     // MARK: - Updates
 
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -1355,6 +1355,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     func toggleHearMyself() {
         guard menuState.mode == .connectedServer else { return }
         connectionController.toggleHearMyself { [weak self] enabled in
+            self?.menuState.setHearMyselfEnabled(enabled)
             let key = enabled ? "shortcuts.hearMyself.announced.on" : "shortcuts.hearMyself.announced.off"
             self?.announceWithVoiceOver(L10n.text(key))
         }
--- a/App/ttaccessible/AppKit/AudioGainControlView.swift
+++ b/App/ttaccessible/AppKit/AudioGainControlView.swift
@@ -112,6 +112,12 @@ final class AudioGainControlView: NSView {
         return true
     }
 
+    override func accessibilityPerformPress() -> Bool {
+        // VoiceOver double-tap resets the gain to unity (0 dB = 50% on the slider),
+        // matching the channel mixer's reset-on-press.
+        resetToZeroAccessibilityAction()
+    }
+
     override func accessibilityChildren() -> [Any]? {
         []
     }
@@ -137,6 +143,12 @@ final class AudioGainControlView: NSView {
     func adjust(by delta: Double) {
         let updated = min(max((slider.doubleValue + delta).rounded(), 0), 100)
         setAndNotify(Self.gainDB(forPercent: updated))
+    }
+
+    /// Nudge one step and return the spoken value (used by the mixer's Cmd+Up/Down).
+    func adjustAndDescribe(up: Bool) -> String {
+        adjust(by: up ? 1 : -1)
+        return Self.format(percent: slider.doubleValue)
     }
 
     func setAndNotify(_ value: Double) {
--- a/App/ttaccessible/AppKit/ChannelMixerCoordinator.swift
+++ b/App/ttaccessible/AppKit/ChannelMixerCoordinator.swift
@@ -1,0 +1,351 @@
+//
+//  ChannelMixerCoordinator.swift
+//  ttaccessible
+//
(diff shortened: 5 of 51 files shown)


math65:
Thanks Rocco — really nice work. I rebased the branch locally on `main` (the #21 commits drop out cleanly, audio-only diff is the 2 commits), it **builds**, and I did a first round of **listening tests**: no audible regressions — no crackling on a 48 kHz→44.1 kHz output, multi-talker mixing sounds clean, live output-device switching no longer freezes (the deadlock cure works), and the mixer pan/mute/solo all behave. The output-bypass architecture clearly does what it promises. 👏

I also ran a careful code review over the audio engine, the SDK lifecycle, the mixer, and the dylib patch. Nothing critical, no audioblock leaks or double-frees, and the RT render callback respects the no-alloc/no-lock contract. A few things I'd like your take on before we ship this as its own beta — happy to take any of these myself, or leave them to you, whatever you prefer. Line numbers are approximate against the rebased branch.

### Higher priority

1. **`patch-sdk-portaudio.py` — the "fails loudly" guard isn't actually in the script.** The docstring promises it survives SDK bumps and fails loudly, and the `nm` symbol lookup does fail if the symbol disappears — but the bytes get overwritten without verifying the original prologue. On the current SDK the prologue is correct (arm64 `sub sp,sp,#0x50`, x86_64 `push rbp;mov rbp`), so today it's fine. But a future SDK that keeps the symbol while changing the prologue would silently produce a corrupt dylib for everyone. Could we add an allowlist check of the original bytes before writing (and `sys.exit(1)` otherwise)? Also the offset math assumes `__TEXT.vmaddr==0 && fileoff==0` — worth asserting that explicitly.

2. **Device preference not re-applied on reconnect.** `appliedInputPreference` / `appliedOutputPreference` aren't reset in `destroyLocked`, while all the other audio-ready flags are. With instance reuse, the dedup guard in `applyAudioPreferences` can see `applied == newPref` after a disconnect→reconnect and silently skip re-initializing the device. Looks like a 2-line fix (reset both to `nil` in teardown).

3. **Mixer keyboard handler does synchronous system-wide AX IPC on every keystroke.** `findFocusedStripUserID()` walks up to ~8 `AXUIElementCopyAttributeValue` calls on `AXUIElementCreateSystemWide()` on the main thread, for most unmodified keys and key-repeats. That's a jank/latency risk under repeat. Could we filter the key first (and bail early if a text field has focus), or cache the focused element (the `VirtualControlView.currentFocused` registry already exists but is currently unused — see also the dead-code note below)?

4. **RT state published without a memory barrier.** In `OutputAudioRenderEngine.startImpl`, `rtPull` / `rtPullCapacity` / `rtPlanePtrs` / `currentGain` are plain `var`s read by the render thread with no fence (unlike the ring/gain cells which do use acquire/release). At first `start` the `AudioOutputUnitStart` acts as a de-facto barrier, but on `switchDevice` the invariant isn't written down. A release fence after the init block + acquire at the top of `render` (or an explicit comment that stop/start provides it) would make it safe-by-construction.

### Medium

- **Persisted pan isn't restored when a user joins.** Volume, media volume and stereo are re-applied on `CMD_USER_JOINED`, but the continuous pan (`userVolumeStore.pan`) isn't pushed to the render engine — so the mixer strip shows the saved pan while the audio plays centered until you touch the slider.
- **Keyboard monitor has no deterministic `stop()`** on disconnect — cleanup relies on the VC's `deinit`. A `stop()` on session close would be more robust.
- **`engineQueue.sync` from the TeamTalk queue** blocks the SDK serial queue during CoreAudio init/start (latency under hot-plug). `async` where the caller doesn't need the result might help.
- **Mix has no limiter** — several loud talkers at once can clip on peaks. Listening was fine for me, just flagging the design.
- **`SoundPlayer`: `applyDeviceLocked()` runs before the `!engine.isRunning` guard** — neutralized today by the `appliedDeviceID` check, but it's the exact AUHAL "set CurrentDevice on a running unit" trap from CLAUDE.md; safer to move it after the guard.

### Two things worth a deliberate decision

- **The input-side silent-switch fallback is still there.** `resolveOutputDevice` now returns `nil` on no match (good), but `resolveCurrentInputDevice` still falls back to `defaultInputDevice ?? devices.first` — that's the known "wrong mic silently selected" bug. The two paths now have inconsistent fallback semantics; want to align them?
- **The user-volume curve changed from piecewise-linear to geometric.** Unit tests still pass (anchors 0/50/100 + round-trip), so no regression — but it's a behavior change and CLAUDE.md still documents piecewise-linear. Intentional? If so I'll update the docs.

### Low / polish
- The whole `VirtualControlView` static registry (`registry`, `currentFocused`, `setAccessibilityFocused`, …) is dead code — never referenced outside the file (and it's `nonisolated(unsafe)` mutable state). Either wire it up to fix #3 above, or drop it.
- VoiceOver `accessibilityRoleDescription` strings ("slider", "button", "channel strip") are hardcoded English while the labels are localized — VO says "channel strip" in French.
- A couple of stale comments (e.g. "restarts on the next muxed block" — it now starts directly; the `PerUserMixSource` "touched only on the TeamTalk serial queue" comment, it's actually `engineQueue`), and an orphan `*.slow-v5.22a-backup` pattern in `.gitignore` that nothing creates.

None of this blocks the listening side — the engine sounds and behaves well. How would you like to handle it? I'm happy to fix all of these myself on the branch, or if you'd rather take them so the audio code stays in your hands that's totally fine too. And do let me know if you disagree with any of the suggestions or read any of them differently — several are judgment calls. Your call. 🙏

rfiorentino1:
Thanks for the thorough pass, Mathieu — and for the listening tests. I took these myself to keep the audio code in my hands; pushed in `7e06c72`. Rundown:

**Higher priority — all addressed**
1. **`patch-sdk-portaudio.py` guard** — now allowlists the known-good unpatched prologue bytes per arch (captured from a fresh v5.22a build) and asserts `__TEXT.vmaddr==0 && fileoff==0` explicitly. If a future SDK keeps `_IsFormatSupported` but changes its prologue, it `sys.exit(1)`s **before writing** instead of corrupting the dylib. Tested: patches the real prologue, idempotent re-run, and bails on a mutated prologue.
2. **Device pref on reconnect** — `appliedInputPreference`/`appliedOutputPreference` now reset to `nil` in `destroyLocked` alongside the other audio-ready flags.
3. **Mixer keyboard AX IPC** — gated `findFocusedStripUserID()` so the system-wide AX walk only runs for keys the mixer acts on (plain arrows + v/p/m/s). Typing, modified keys and unrelated shortcuts pass straight through. The real mixer interactions are byte-for-byte unchanged.
4. **RT barrier** — added a release fence after the render-state block in `startImpl` and an acquire at the top of `render()`, matching the ring's existing acquire/release. Makes the `switchDevice` publication safe-by-construction instead of leaning on `AudioOutputUnitStart`'s implicit barrier.

**The two judgment calls**
- **Input fallback** — aligned: `resolveCurrentInputDevice` now returns `nil` on no-match for an *explicitly chosen* device, like `resolveOutputDevice`. The init path already throws `deviceUnavailable`, so the wrong mic is never silently substituted. I left the `usesSystemDefault` branch's fallback intact on purpose — that path's intent really is "follow the OS."
- **Volume curve** — intentional, please keep it. The geometric/dB-linear mapping is deliberate (the code comment explains why: a plain linear-gain map made the top half brutal — 50→51% was ~+4 dB while 99→100% barely moved). Anchors + round-trip still pass. Go ahead and update the docs to say geometric.

**Medium / polish — done**
- Persisted **pan restored on join** (both join paths). `muted:false` is the engine default; an active solo is re-applied right after by the coordinator's `reapplySolo()` on the next session update.
- **SoundPlayer** — `applyDeviceLocked()` moved after the `!engine.isRunning` guard, so `CurrentDevice` is only set on a stopped unit.
- Dropped the dead `VirtualControlView` registry (I gated the key in #3 rather than wiring it).
- Localized the VO **role descriptions** (en/fr: `curseur` / `bouton` / `tranche de console`).
- Stale comments fixed (engine starts directly; `PerUserMixSource` is `engineQueue`-only).

**Left deliberately**
- **`*.slow-v5.22a-backup` in `.gitignore`** — left it. `Vendor/TeamTalk/.gitignore`'s `*.dylib` doesn't match a `…dylib.slow-v5.22a-backup` name, so that root line is the only thing keeping a local 72 MB backup out of `git status`. Cheap insurance against committing a big binary — I'd rather keep it.
- **No limiter / `engineQueue.sync→async` / keyboard `stop()` on close** — left as-is for now. The limiter and the sync→async are design changes I'd want to measure before touching, and the keyboard monitor's `deinit` cleanup works today. Happy to revisit any of them if you feel strongly.

Disagree with any of the calls above? Easy to adjust.

math65:
Merged, and it shipped today as **1.7.0-beta.7** — notarized and live, and I posted it on the AppleVis beta-testers thread so a few people will be putting the mixer through its paces. Thank you again for all of this, and for taking the review fixes yourself; the patch-script guard and the RT fence are exactly right.

And yes — please go ahead with that separate PR for the in-app discoverability of the mixer shortcuts (help tags + a VoiceOver-friendly shortcuts reference). That's the one real gap left, and since the shortcuts currently aren't written down anywhere in the UI, it'd be a great addition. No rush; whenever suits you.

The other items you left deliberately (limiter, the sync→async, the keyboard stop() on close) I'm happy to leave as-is for now too — agreed they're worth measuring before touching.

PR #27, opened 2026-07-07
Guarantee the PortAudio probe patch is applied at build time

rfiorentino1:
The `IsFormatSupported` patch that removes PortAudio's ~13 s startup device probe (added in beta.7) is only applied inside `download-sdk.sh`, and only on a *fresh* download.

`libTeamTalk5.dylib` is a gitignored artifact, so a build machine that already has an unpatched cached copy (downloaded before the patch existed) never gets it re-applied — nothing at build time ensures it — and ships the slow-connect build. The distributed **v1.7.0-beta.9** binary embeds an unpatched dylib for exactly this reason: connecting to a channel takes ~13 s on larger device setups instead of being near-instant.

**Fix**
- Add a `Patch TeamTalk SDK` run-script build phase that runs the existing idempotent, fail-loud patcher on the vendored dylib before it is linked/embedded — so no build can ship unpatched regardless of how the dylib got onto the machine.
- Decouple the patch in `download-sdk.sh` so re-running it patches an existing dylib instead of skipping entirely when the user declines the re-download.

No product code changes; the dylib stays out of the repo.

**Verified:** built from a deliberately-unpatched vendored dylib and confirmed the embedded dylib in the resulting `.app` comes out patched. Also confirmed patching a distributed beta.9 in place makes connect near-instant.

### DIFF
--- a/App/ttaccessible.xcodeproj/project.pbxproj
+++ b/App/ttaccessible.xcodeproj/project.pbxproj
@@ -38,6 +38,24 @@
 		};
 /* End PBXCopyFilesBuildPhase section */
 
+/* Begin PBXShellScriptBuildPhase section */
+		94A4A5FA3B941A8E00C01234 /* Patch TeamTalk SDK (PortAudio probe) */ = {
+			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Patch TeamTalk SDK (PortAudio probe)";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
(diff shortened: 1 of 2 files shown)


math65:
Merged and shipped as **v1.7.0-beta.10** (build 41) 🎉 — notarized (Accepted), stapled, and live on the beta channel.

Thank you for this, Rocco. This was a genuinely important catch: I ran your patcher against my own release machine's cached `libTeamTalk5.dylib` and it came back **unpatched** — so beta.9 really did ship the ~13 s slow-connect binary, exactly as you diagnosed. The build-phase guarantee means that can't happen again regardless of how the dylib landed on the machine, and beta.10 is the first build where the patch is actually embedded end-to-end (confirmed: the phase reported the dylib patched before embed).

Clean, well-reasoned, and no product-code risk. Really appreciate you chasing this down. 🙏

Release: https://github.com/math65/ttaccessible/releases/tag/v1.7.0-beta.10

Issue #4, opened 2026-05-17
Cross-project introduction: TeamTalk VO Client + iOS app ideas


fla-rion:
Hi @math65,

I'm Florian Lichteblau, the developer of **TeamTalk VO Client** — a Python-based TeamTalk client focused on accessibility (VoiceOver, Braille display, screen reader support) for macOS and Windows. I came across ttaccessible and was genuinely impressed by the architecture, especially the dual audio engine approach and the native VoiceOver integration without runtime patching.

**Our project:** https://github.com/fla-rion/TeamTalk-VO-Client

---

## What our client currently offers

The Python client has grown quite large over time. Here's a rough feature overview:

**Core**
- Full server/channel/user management, multi-server profiles, .tt file / tt:// URL support
- Channel and private chat with history, search, export (txt/html)
- File transfer, admin tools (ban list, user accounts, server broadcast, server stats)
- Auto-reconnect, auto-join per server

**Audio**
- PTT + Voice Activation, AGC, noise suppression, echo cancellation (SDK-level)
- Per-user volume + stereo position, media file mute/volume separate from voice
- Equalizer presets, per-server audio profiles
- Global hotkeys (PTT + mute-all) — macOS via NSEvent monitor, Windows via Win32 polling

**Accessibility**
- VoiceOver support on macOS (via ObjC runtime patching — your native approach is clearly better)
- Braille display support (compact label mode)
- NVDA support on Windows
- Configurable TTS via espeak-ng with per-context rate/voice overrides (chat vs. system vs. channel events), pronunciation dictionary, interrupt mode

**Automation / AI**
- Macro manager (text macros + actions, scheduled macros)
- Trigger rules (event → action)
- Auto-reply manager
- Chat translation (DeepL/LibreTranslate)
- AI channel summary on join (Gemini API — summarizes last 30 min of chat)
- Mute scheduler (time-based mute windows)

**Misc**
- Media streaming: YouTube/yt-dlp, SoundCloud, Twitch, Bandcamp, Vimeo, Mixcloud, Webradio, Podcast, local file, playlist
- Recording (WAV/MP3, scheduled recordings)
- Plugin system (Python plugins loaded at runtime)
- Webhook manager, HTTP API server
- Bookmark manager, channel notes, user notes
- Saved messages, offline message queue
- Settings backup/restore, server list import/export
- Chat search, user watcher, online users dialog
- Startup profiler, health check, usage analytics

---

## Our interest in your project

[...]

Issue #2, opened 2026-04-14
refactoring to avoid god objects and rely on clean separation of concerns


xogium:
Hi!
I checked out the code yesterday, and the first thing I noticed is how... Monolithic it all is. No offense to you, but Claude clearly has no reflexe to actually create untangled code.

It's quite a mess, some parts of the program know way, way too much about other parts without it being needed. It is difficult to maintain such an untangled codebase, both for AI and for humans, so I am proposing to work on a pull request to untangle everything, to have a clean separation of concerns and no god objects.

Of course, before I actually submit this work, I'd do a comprehensive bunch of testing, manual/runtime as well as automated testing if possible, to ensure that absolutely nothing breaks.

Is this something that would be of interest?

It will also make it much easier in the future to add features and fix bugs in existing features without accidentally affecting whole building blocks of the code that are unrelated.

math65:
Hey,

Sorry about the mess! I want to be transparent — I do use AI tooling for most of the implementation work, and despite my efforts to refactor here and there, things clearly got tangled over time. It's a lesson learned for me in terms of keeping tighter architectural oversight.

Really grateful you're willing to tackle this. Don't hesitate to reach out if you have any questions while working on it!

xogium:
Hi!

Just a small heads up, that I am working on a gigantic refactor branch at the moment. Just to let you know that I haven't forgoten about this at all.

I also use AI, for what it's worth, however, but mostly because I simply cannot grasp the coding itself, but I know what needs to be done, if that makes any kind of sense. My priority is to avoid breaking things at all costs, so I will take my own build for a spin for a couple weeks before I actually open a pull request. We absolutely can't break it. Not now people are actively using it.

I'm also working on a suie of unit tests.

math65:
Hi!

Thanks for the heads up, that sounds great! No rush at all, take all the time you need. Totally agree we can't break things now that people are relying on it.

And same here regarding AI, no judgment at all! Feel free to add or change whatever you think is needed, you have full freedom on this. Really looking forward to seeing the refactor and the unit tests.

Cheers!

PR #29, opened 2026-07-17
Device audio streaming, admin file-storage controls, upload fix, macOS 12 support, UI polish

rfiorentino1:
This is a batch of features and fixes from the last week of daily use and tester feedback. Commits are self-contained and reviewable one by one; happy to split any of this into separate PRs if you prefer.

## Stream Audio Device (new feature)

"Stream Audio Device…" (Cmd+Opt+A) next to Stream File/URL: broadcasts a live audio input device to the channel as a media stream. The SDK has no raw-PCM-to-media API (`TT_InsertAudioBlock` is voice-only), so the device is captured via CoreAudio and served as an endless AAC/ADTS stream (Apple's native encoder, 192 kbps) on a loopback HTTP server that `TT_StartStreamingMediaFileToChannelEx` consumes like any URL. AAC rather than raw PCM because FFmpeg's open-time stream analysis consumes ~50 packets that become permanent broadcast latency — 85 ms PCM packets cost ~4.3 s, 21 ms AAC frames ~1.1 s (measured against the SDK's own FFmpeg), for ~2 s total latency (field-verified end to end, down from 6+ s with PCM). The server paces output to the wall clock with a 300 ms buffer and pads silence when capture stalls, so a quiet or unplugged device never ends the broadcast. Self-monitoring is opt-in per stream (checkbox, off by default); file/URL streams keep their existing always-monitor behavior. Covered by an integration test (real capture + SDK probe + pacing assertion).

## Fixes

- **Held media-volume/seek keys kept adjusting after release**: each key repeat queued an SDK call; on machines where those run slower than the repeat rate the backlog kept draining after release. Both controls now coalesce to latest-value-wins (`CoalescedRequest`, unit-tested). Also reshaped to dodge a Swift 6.3.2 release-optimizer crash (EarlyPerfInliner on a generic class deinit).
- **False "channel does not have enough storage" on upload**: removed the client-side quota pre-check. Verified against a live server that its math diverges from server enforcement (admins bypass the channel quota, and the server's accounting is its own). A genuine rejection still surfaces the same localized message via `CMDERR_MAX_DISKUSAGE_EXCEEDED`.
- **Intercept sound fired on every login**: the post-login sync re-establishes peers' pre-existing subscriptions across several user updates and each looked like a fresh transition. Subscription state is now recorded silently inside the existing login-suppression window; mid-session changes still alert.

## Admin tools

- Channel create/edit gains a **disk quota** field with a KB/MB/GB unit picker (stored as bytes in `nDiskQuota`; opens in the largest clean unit; switching units converts the displayed number; untouched values keep their exact bytes).
- Server Properties now shows the remaining `ServerProperties` fields: editable TCP/UDP ports plus read-only server/protocol versions.

## macOS 12 support — your call

Deployment target lowered 14.0 → 12.0 (commit `b57bf43`). Everything in the dependency chain already ran on 12; the code changes are: an `onChange` compatibility wrapper, the User command menu's visibility managed at the AppKit layer (CommandsBuilder has no conditionals before macOS 13, and the availability-split alternative relies on an `Optional: Commands` conformance that doesn't exist at runtime on 12), and `loadView` overrides for programmatic view controllers (pre-14 AppKit throws without a nib; 14+ silently creates an empty view — worth having regardless). Field-tested on macOS 12.7.6 (Intel). **If you'd rather stay on 14, say the word and I'll drop the deployment-target commit — the loadView/onChange hardening is worth keeping either way.**

## UI polish

- User Accounts (Cmd+Shift+U) gains a "Nickname (online)" column showing the current nickname of whoever is logged in with each account.
- All auxiliary windows (accounts, users, bans, messages, files, info, stats, profiles, editor, feedback, preferences) close on Escape via a shared `EscapeClosableWindow`; text-field editing still consumes Escape first, and the main session window intentionally keeps requiring Cmd+W.
- `build.sh`: tag the English `releaseNotesLink` with `xml:lang="en"` so Sparkle stops warning when the French sibling is present.

All strings localized (en + fr), full test suite passing, notarized builds of this exact content field-tested on macOS 12 and 26.


math65:
Thanks for this — great batch, and the device-streaming approach (loopback AAC/ADTS so the SDK consumes it like any URL) is clever. A few things before I merge.

**macOS 12: yes, let's keep it.** Since it's field-tested on 12.7.6 and the `loadView`/`onChange` hardening is worth having anyway, keep the deployment-target commit.

I did a close pass on `AudioDeviceStreamSource`. One blocker and a few things worth tightening:

**1. Blocker — crash in `stop()` (dictionary mutated while enumerating).**
`for connection in connections.values { connection.cancel() }` — `cancel()` → `finish()` → `onFinish(self)` runs synchronously and does `connections.removeValue(...)`, mutating the dict mid-enumeration. This traps essentially every time a stream is stopped with a live connection (the SDK's FFmpeg reader is always one). Snapshot into `Array(connections.values)` before the loop, or clear the dict first.

**2. Real-time thread — heap allocation + lock in the AUHAL callback.**
`handleInput()` runs on the CoreAudio RT thread and allocates `stereo`, allocates the resampled array, then takes `PCMRing`'s `NSLock`. This diverges from the pre-allocated / zero-alloc pattern we rely on elsewhere (documented in CLAUDE.md for the AUHAL path) and risks priority inversion / dropouts under load. Could you pre-allocate the scratch buffers and avoid locking on the RT side?

**3. `startServer()` blocks the serial SDK queue up to 5 s.**
It's called synchronously on the shared TeamTalk queue and waits on `readySemaphore` (5 s timeout). Normally instant, but a slow `.ready` freezes the whole SDK event loop. Worth making the wait non-blocking or bounded much tighter.

**4. Nit — accounts list does one `TT_GetUserByUsername` per account.**
On a server with many registered accounts, opening the User Accounts window runs that many serial lookups. A single username→nickname map built from the connected users would avoid it.

**5. Nit — `stopped` / `listener` are written from the TeamTalk queue** although they're documented as serverQueue-guarded (and `listener` is nil'd from serverQueue in the failed handler). Minor data race; set initial state before `start()` and mutate only on serverQueue.

#1 is the only merge blocker; #2 and #3 I'd like fixed since it's real-time audio, #4/#5 are your call. Everything else — coalescing, quota field, Escape-to-close, the login-suppression fix — looks solid. Thanks!

rfiorentino1:
Thanks for the close read — exactly the pass I wanted on the RT path. All six items are pushed; one heads-up first: **the device-streaming transport changed since your review.**

**Transport: AAC/ADTS → Ogg Opus.** The AAC loopback still leaned on the SDK's FFmpeg reader doing `find_stream_info`, which stalls under the SDK audio lock on stream start (the freeze) and replays the buffered probe as latency. Serving **Ogg Opus** instead gives the reader instant stream info — no probe stall — so both the start-freeze and the added latency go away at the root, while it stays a genuinely separate media stream (not voice). It uses the `opus_*` symbols the vendored `libTeamTalk5.dylib` already exports, via a thin non-variadic C shim — no new dependency. Files: `OggOpusStreamEncoder.swift`, `OpusShim.{c,h}`.

On your points:

**1 (blocker — `stop()` crash):** Fixed. Snapshot into `Array(connections.values)` and clear the dict before cancelling, so the synchronous `onFinish` → `removeValue` can't mutate mid-enumeration.

**2 (RT-thread alloc + lock):** Fixed. The input callback is now lock-free and allocation-free. `PCMRing` is single-producer/lock-free — the producer copies into a raw buffer and publishes the absolute write cursor with a release store; readers acquire-load it and seqlock-validate the copy, so no lock ever touches the RT thread (removes the `NSLock` and its priority-inversion risk). The stereo-map and resample scratch buffers are both pre-allocated, and the muted path zeroes the scratch in place — no per-callback allocation. (The atomic is a tiny C11 shim, `AtomicU64`, since `Synchronization.Atomic` needs macOS 15 and the target is 12.)

**3 (`startServer()` blocking the SDK queue):** Fixed. The device open + server start (the `readySemaphore` wait) now run on a background `userInitiated` queue; only the post-ready bookkeeping hops back onto the shared TeamTalk queue, so a slow `.ready` can't freeze the SDK event loop.

**4 (accounts N× `TT_GetUserByUsername`):** Fixed. The listing now builds a single lowercased-username → nickname map once from `TT_GetServerUsers` (first login wins, matching the old per-account behaviour) and resolves each account's online nickname from it — no per-account SDK call.

**5 (`stopped`/`listener` race):** Tightened — dropped the redundant `stopped` write off the TeamTalk queue; each stream gets a fresh source so initial state is set before `start()`, and `listener` is only touched on `serverQueue`.

That's all five plus the transport.

PR #30, opened 2026-07-17
VoiceOver: Channel Mixer per-stream pan + stereo fix, region announcement, mute-state labels; recording ⌘R/⌘⇧R modes

rfiorentino1:
Four accessibility/usability improvements, all building on features already in v1.7.0:

**Channel Mixer per-stream pan + stereo handling** — Voice and media pan are now independent, mirroring the split voice/media volume controls. Plain ←/→ (and `p` to announce / double-`p` to reset) drive **voice** pan; ⌘←/→ (and ⌘`p` / double to reset) drive **media** pan — strip-gated like the existing ⌘↑/↓ media-volume keys, so off a strip they pass through. A second "Media pan" slider is added and the strip is grouped by stream (voice vol, voice pan, media vol, media pan, mute, solo). Voice pan reuses the existing stored key (old pans carry forward as voice); media pan uses a new one. Two audio fixes come with it: a stereo sender is downmixed to mono *before* the pan gains when panned off center, so it repositions cleanly instead of just fading one channel's content (a centered stereo sender stays true stereo); and pan announces "Center" vs "Stereo" from the actual L/R content (per-block side/mid energy, latched), **not** the SDK's codec-driven `AudioBlock.nChannels` — a stereo-codec channel decodes even a mono phone mic to two identical channels, which was reading as false "Stereo". Judged per stream, so voice and media report independently.

**Channel Mixer region announcement (⌘5)** — Focusing the mixer now speaks "Channel Mixer" before the landed strip/area, instead of the focus change swallowing the region name. Done by prepending a one-shot prefix to the focused element's label so both are spoken in one utterance.

**Toolbar mute-state labels** — The Microphone and master-mute toolbar items now fold their muted/unmuted state into the label VoiceOver reads (mic: "Microphone, muted/unmuted"; master relabeled from "Mute/Unmute" to "Master, muted/unmuted"). Pressing the mic control suppresses the now-redundant spoken status (VoiceOver already reads the state-bearing label); the keyboard shortcut/menu path keeps it.

**Recording ⌘R vs ⌘⇧R** — ⌘R always records a single mixed file; ⌘⇧R records using the preference mode (separate files, or both). The recording-mode preference now only offers separate/both (single is always ⌘R); a legacy "single" value migrates to "both" (a superset that still produces the single muxed file). While recording, a single Stop item is shown (⌘R), and ⌘⇧R also stops. Start announcements delineate mode ("Recording started: …", "Stems recording started", "… plus stems").

French localizations included.


math65:
Thanks for this — clean batch, and the stereo-aware pan handling (content-derived Center/Stereo + mono downmix before the pan gains) is a really nice touch.

One point I'd like your call on before merging, plus a few minors.

### Needs a decision — auto-restart replays the *preference* mode, not the mode actually in use

`AppDelegate.startRecordingToFolder(_:mode:)` stores `activeRecordingMode = mode`, but the auto-restart path re-reads the preference instead:

```swift
// AppDelegate.swift (~L2199, TeamTalkConnectionControllerDelegate auto-restart)
startRecordingToFolder(folderURL, mode: preferencesStore.preferences.recordingMode)
```

So if I start a **single-file** recording with ⌘R (`activeRecordingMode == 1`) and then a channel change / reconnect / relaunch triggers auto-restart, it comes back in the preference mode (separate or both) and silently starts writing per-user stems I never asked for. Feels like auto-restart should restore `activeRecordingMode` when a recording was active, rather than the stored preference — but since you own the recording UX here, I'll leave the call to you.

### Minor / non-blocking

- **`AudioBlockPump.blockIsStereo` runs every block even after latching.** In `updateStereoJudgement`, `channels >= 2 && blockIsStereo(...)` evaluates the full mid/side energy scan on every block; once a key is latched to stereo it's pure waste on the pump's 10 ms queue (matters most in Debug). Gating it behind `sawStereoByKey[engineKey] != true` skips the scan once proven.
- **Region prefix cleared by a fixed 1.5 s timer.** `applyRegionPrefix` clears via `asyncAfter(1.5s)`; two quick ⌘5 in a row schedule two timers and the first can clear the prefix mid-read of the second focus. Clearing on the next focus event instead of a timer would be tighter (very minor).
- **Visible SwiftUI faders never say "Stereo".** The on-screen faders call `panDescription($0)` without the channel count, so they show "Center" while VoiceOver says "Stereo". Cosmetic (the visible UI is accessibilityHidden), just noting the divergence.

On the legacy `recordingMode 1 → 3` migration: that one's clearly intentional given the ⌘R/⌘⇧R split — I'll just call it out in the release notes so existing "single file" users know the toolbar button now records both.

Everything else looks good to me.

rfiorentino1:
Thanks for the careful read — pushed a follow-up (a0754a8) addressing all four.

**Auto-restart mode** — agreed, it should restore what was actually in use. The root cause is that `recordingMode` is clamped to 2/3 by `clampRecordingMode`, so a single-file ⌘R recording (mode 1) could never round-trip through the preference and always came back writing stems. Added a `lastActiveRecordingMode` preference stored raw (1/2/3) alongside `lastRecordingWasActive`, set when a recording starts, and the auto-restart path now replays that — falling back to the preference only if it's somehow unset. Covers channel change, reconnect, and relaunch.

**`blockIsStereo` per block** — gated behind `sawStereoByKey[engineKey] != true`, so once a key latches to stereo the mid/side scan is skipped entirely.

**Region prefix timer** — made the delayed clear a cancellable `DispatchWorkItem`; a second ⌘5 now cancels the first announcement's pending clear, so it can't blank the prefix mid-read. I kept the short window itself (it's what stops ordinary navigation back to the element from repeating the prefix) — the race was the bug, not the timer.

**Visible faders** — they now call the same stereo-aware `voicePanDescription`/`mediaPanDescription` the VoiceOver overlay uses, so the on-screen text and VO agree ("Stereo" vs "Center").

On the `recordingMode 1 → 3` migration — sounds good, I'll leave the release-notes callout to you.

PR #31, opened 2026-07-19
Push-to-talk: custom hotkey engine, "Both" mic mode, per-hotkey global/local scope

rfiorentino1:
## What

Reworks push-to-talk and the ⌘⇧A mute toggle:

- **New hotkey engine** (`HotkeyMonitor`) that works under the App Sandbox. Each binding is either app-focused or global. Global uses a listen-only `CGEventTap` (Input Monitoring); app-focused uses an `NSEvent` local monitor. The PTT monitor is only active in push-to-talk / both modes, so it never captures the key in always-on mode.
- **New "Both" microphone mode**: ⌘⇧A toggles an always-on gate, holding the PTT key transmits regardless of the gate, and releasing goes silent. The mic engine is kept hot while in a channel so PTT is instant; the gate is a lightweight software flag.
- **Accessible key recorder** — a real button, not a search field — that ignores VoiceOver's Control-Option activation modifiers when recording a chord.
- **Preferences**: a microphone-mode popup (with Both), a push-to-talk key field, and per-hotkey global/local toggles for both PTT and ⌘⇧A.
- ⌘⇧A is ignored while Finder is frontmost.

## Notes

- Independent of #29 (device streaming). It shares one small helper with #30 — the `announceStatus:` parameter on `performToggleMicrophoneShortcut`, from #30's VoiceOver mute-state work — folded in here so this PR stands alone; whichever of #30 / this merges second gets a trivial one-line overlap on that method.
- Sandbox caveat worth flagging: a global binding can *detect* the key (Input Monitoring) but can't *swallow* it — swallowing needs Accessibility, which the sandbox blocks. So a global PTT/⌘⇧A fires our action but the key still reaches the focused app; app-focused bindings swallow normally.
- Built clean (Debug, macOS 12 target).


math65:
Really nice work — the sandbox-safe hotkey engine and the "Both" mode are a clean design, and the accessible key recorder (real button, VO-modifier filtering) is a genuine improvement over the old library. I reviewed the whole diff carefully; the concept is solid, so everything below is about edge cases in the transmit state machine and the global-hotkey path. Grouped by priority.

### Blocking — mic can transmit without the user intending it

1. **Mode switch leaves the gate/engine state stale (`+Audio.swift`, `handleMicrophoneModeChanged`).** Leaving a muted `.both` (engine hot, `bothGateOpen == false`) for `.alwaysOn` never clears `voiceTransmissionEnabled`, so the mic goes live with no user action. Symmetric case: open the `.both` gate → switch to `.pushToTalk` (`bothGateOpen` left `true`) → back to `.both`; `armBothModeEngineIfNeededLocked` early-returns on `voiceTransmissionEnabled != false` and never re-closes the gate, so it resumes transmitting. Suggest: on every mode change, normalize state explicitly — reset `bothGateOpen = false` and re-derive `voiceTransmissionEnabled` from the target mode rather than leaving whatever the previous mode set.

2. **Local-scope PTT release is lost when focus leaves (`HotkeyMonitor.swift`).** Hold the key, Cmd-Tab away: the `keyUp` goes to the other app, the local `NSEvent` monitor never sees it, `pushToTalkPressed` stays `true`, mic keeps transmitting. There's no `applicationWillResignActive` cleanup. Suggest: on `resignActive` (or window resign key), force `setPressed(false)` / `setPushToTalkPressed(false)`.

3. **Global `⌘⇧A` is hardcoded to `keyCode 0`, which is layout-dependent (`AppDelegate.swift`).** The menu uses `.keyboardShortcut("a", …)` (character-based). On AZERTY, `keyCode 0` is the **Q** key — so global `⌘⇧A` does nothing and `⌘⇧Q` toggles the mic. Best fix ties into #14 below (make it a configurable binding); minimal fix is to resolve the keyCode for the current layout instead of hardcoding 0.

4. **No migration of the old push-to-talk binding.** Removing `KeyboardShortcuts` drops its `KeyboardShortcuts_pushToTalk` value; `pushToTalkKey` defaults to `nil`, so a user who had PTT configured silently falls back to always-on transmission (only hinted at by the "No hotkey configured" warning). Suggest: one-time read of the old UserDefaults key into a `HotkeyBinding` on first launch, or at minimum a release-note callout that PTT must be re-set.

### Should-fix — correctness

5. **A single-character PTT key is swallowed app-wide in local scope (`HotkeyMonitor.handleNSEvent`).** The recorder invites single keys, but then that character can't be typed into the chat field. Consider not swallowing (or not matching) a bare printable key while a text field is first responder.

6. **The global tap drops `.numericPad` (`HotkeyMonitor.modifierFlags(from:)`).** An arrow/keypad PTT key has `binding.modifiers == [.numericPad]` but the CG conversion never sets it, so `mods == binding.modifiers` is always false → the key works locally but never globally. Either map `.numericPad` from `CGEventFlags`, or strip it from `binding.modifiers` at record time so both paths agree.

7. **Self "talking" indicator goes stale on PTT (`+Audio.swift`, `setPushToTalkPressed`).** The snapshot's own `isTalking` now comes from `isEffectivelyTransmittingLocked`, which flips with `pushToTalkPressed`, but `publishSessionLocked` only runs in the both-mode gate-close branch. In `.pushToTalk`/`.both`, pressing/releasing changes transmission but VoiceOver's state doesn't refresh. Suggest: publish on any `pushToTalkPressed` transition.

8. **`announceStatus` is honored only in the `.both` branch of `toggleMicrophone` (`ConnectedServerViewController.swift`).** The `.alwaysOn`/`.pushToTalk` branches call `announce(…)` unconditionally, so a mic-button click is silent in one mode and announced in the others. Thread `announceStatus` through both branches.

9. **`willOpen` read from the stale main-thread snapshot (`toggleMicrophone`).** Two quick `⌘⇧A` can announce "enabled" while the gate actually closed. `toggleBothModeGate` already knows the authoritative result — return it (e.g. `Result<Bool, Error>` = gate-now-open) and announce / gate mic-access off that instead of `session.voiceTransmissionEnabled`.

10. **`currentMicrophoneMode` reads the `@Published` `preferences` struct from the serial queue in hot paths.** That's a cross-thread read of a non-`Sendable` value (data race with `mutate` on main) *and* a whole-`AppPreferences` copy per audio chunk on the <0.2%-CPU insert path. Cache the mode in a plain field updated in the mode sink.

11. **Mute `isPressed` can stick (`handleTapEvent`).** If the app becomes active between `⌘⇧A` down and up, the `globalOnlyWhenInactive && NSApp.isActive` guard eats the `keyUp`, so the next background toggle hits `guard isPressed == false` and is silently missed. Let the disable/keyUp bookkeeping run before the active-app guard.

12. **`stop()` doesn't invalidate the mach port.** `CGEvent.tapEnable(false)` + dropping the refs doesn't release the `CFMachPort`; across repeated reconfigure cycles the receive right can linger. Add `CFMachPortInvalidate(tap)` before clearing.

### Quality / design (non-blocking)

13. **Transmit-gate switch is duplicated** — `insertAdvancedMicrophoneAudioChunkLocked` re-derives the exact switch `isEffectivelyTransmittingLocked` computes (the doc comment even says "Mirrors…"). Collapse the insert guard to `guard isEffectivelyTransmittingLocked, inChannel else …`.

14. **Make `⌘⇧A` a real configurable `HotkeyBinding`** reusing the recorder/monitor infra, instead of a hardcoded chord + per-app ignore list (`com.apple.finder`). Since the tap is listen-only, `⌘⇧A` currently double-fires in *every* app that uses it, not just Finder — the ignore list is whack-a-mole. Making it user-configurable removes the hardcode, the Finder special-case, and fixes the AZERTY bug (#3) in one move.

15. **`pushToTalkShortcutResolver` is now dead indirection** — the controller owns `preferencesStore` and reads it directly elsewhere, so the injected closure can be replaced by a direct `preferencesStore.preferences.pushToTalkKey?.isValid ?? false` at the two call sites, removing a launch-ordering footgun.

One logistics note: this shares the `announceStatus:` line with #30, which is now merged to `main`, so you'll hit that trivial one-line overlap on rebase.

Thanks again — happy to talk through any of these, especially the state-machine normalization in #1.

rfiorentino1:
Thanks for the thorough review — all 15 addressed, and the branch is rebased onto main (the `announceStatus` overlap with #30 resolved in the rebase; that also turned out to fix #8, since the merged branches already honor the flag).

The blocking items are in the second commit: mode changes now normalize the transmit state exactly as you suggested (reset both flags, re-derive `voiceTransmissionEnabled` from whether the old mode's user-facing gate was open), losing active status releases a local-scope PTT hold, and I took the #14 route for #3 — the global mute hotkey is now a recordable `HotkeyBinding` (default resolves "A" through `UCKeyTranslate` for the current layout), which also removed the Finder ignore list. The old KeyboardShortcuts key migrates once on launch, marked so a later manual clear isn't re-imported.

On #9 I went one step further than suggested: `toggleBothModeGate` now returns gate-now-open and the caller requests mic access unconditionally (instant no-op when already authorized), so the stale `willOpen` is gone entirely. Happy to adjust any of it.

PR #32, opened 2026-07-22
Add language preference and first-launch selection dialog

GruiaChiscop:
## Summary

Lets users choose the app's display language instead of always following the
system locale.

- New **Language** picker in Preferences > General (System Default / English
  / French), backed by `AppLanguagePreference` and persisted in
  `AppPreferences`.
- `L10n.configure(languagePreference:)` swaps `L10n.text`/`L10n.format` to the
  chosen locale's bundle at launch and whenever the preference changes.
- A one-time dialog on first launch asks new users to pick a language before
  the rest of the UI appears, instead of only surfacing the setting later in
  Preferences. Guarded by `hasChosenInitialLanguage` so it never shows again
  once answered.
- Help text in Preferences clarifies that switching languages fully applies
  after restarting — the currently open Preferences pane updates live, but
  menu bar commands and other windows don't re-render until relaunch.

## Notes for review

- `L10n.text`/`L10n.format` are called from several non-main-thread contexts
  (e.g. `OutputAudioRenderEngine`, `TeamTalkConnectionController+Audio`, for
  status/error strings). The override-bundle lookup is protected by an
  `NSLock` around both the read and the (rare) write in `configure(...)`,
  since it was previously an unsynchronized `nonisolated(unsafe)` static var.

## Test plan

- [x] Builds clean (Debug, macOS 12 deployment target)
- [x] Verified the bundle-swap mechanism resolves the correct localized
      strings for System Default / English / French directly against the
      built app bundle
- [x] Confirmed `L10n.configure` runs during `AppPreferencesStore.init()`,
      before any UI renders — no flash of the wrong language on launch
- [x] Confirmed the first-launch dialog appears once and doesn't reappear on
      subsequent launches
- [x] Manual pass with VoiceOver on the first-launch dialog and the
      Preferences picker (accessibility of `NSAlert` + `NSPopUpButton` vs.
      the SwiftUI picker)

### DIFF
--- a/App/ttaccessible-Info.plist
+++ b/App/ttaccessible-Info.plist
@@ -4,8 +4,8 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
-	<key>LSApplicationCategoryType</key>
-	<string>public.app-category.social-networking</string>
+	<key>CFBundleDisplayName</key>
+	<string>tt-Accessible</string>
 	<key>CFBundleDocumentTypes</key>
 	<array>
 		<dict>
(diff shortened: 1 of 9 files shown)


math65 (commented):
Thanks for this — and welcome! It's a clean, well-scoped first contribution. It fits the codebase conventions nicely (the `decodeIfPresent ?? default` Codable migration, the `mutate {}` store pattern, the SwiftUI accessibility pattern of a hidden label + explicit `accessibilityLabel`), both locales are updated, and the French uses proper vouvoiement. I especially appreciated the second commit fixing the data race you introduced yourself.

I built it locally (Debug, macOS 12 deployment target) — **builds clean**. I also verified the `overrideBundle` lock is safe: `L10n.text`/`format` are not actually called from the CoreAudio render callback (the one hit at `OutputAudioRenderEngine.swift:799` is `errorDescription`, an error path; the others run on the TeamTalk serial queue), so the `NSLock` doesn't sit on a real-time thread. Good instinct adding it regardless.

Only minor things, no blockers:

### 🟡 1. First-launch dialog accessibility — the one that matters most here
This app is VoiceOver-first, and this dialog is the very first thing a new blind user meets. The `NSPopUpButton` (`AppDelegate.swift`, in `promptInitialLanguageIfNeeded`) has no explicit accessibility label, so VoiceOver announces "System Default, pop up button" without saying what it selects. Could you add `popUp.setAccessibilityLabel(L10n.text("preferences.general.language.label"))` and complete the VoiceOver pass you left unchecked in the test plan (the `NSAlert` + `NSPopUpButton` combo)? That last checkbox is the important one for us.

### 🟡 2. Existing users will see the dialog once after upgrading
`hasChosenInitialLanguage` defaults to `false` when the key is absent, so every current user gets the first-launch language modal on their next update — not just new installs. It's harmless (default = System = their current behavior), but it's a little intrusive for an established user base. Your call whether that's acceptable; if not, one option is to only prompt when the preferences file didn't exist at all (truly fresh install).

### ⚪ 3. Nit — no development-language fallback on a missing key
With `overrideBundle` pointing at a standalone `en.lproj`/`fr.lproj`, `localizedString(forKey:value:nil)` returns the raw key if it's missing from that language, rather than falling back through the main-bundle chain. Low risk given the "both locales always updated together" convention — just something to be aware of.

---

How would you like to handle the fixes? Happy either way:
- **You take them** — they're small and it keeps the PR yours; or
- **I take them on your branch** (maintainer-edit is enabled) if you'd rather not spin back up.

Item 1 is the only one I'd consider needed before merge; 2 and 3 are a decision and a note. Thanks again for a solid contribution.

GruiaChiscop:
@math65 I checked your observations and corrected them. Yes, I forgot to assign a label to the NSPopupButton, I'll push the commit.

math65:
Merged — thanks again, this is a solid first contribution and a genuinely useful feature. 🎉

The VoiceOver label on the first-launch popup is exactly right.

One small heads-up: the same commit also commented out the `.accessibilityLabel` on the per-event sound toggles in `PreferencesSoundsView.swift` (probably a stray local change). Since the visible `Text` there is `accessibilityHidden`, that left each switch with no VoiceOver name, so I restored it in a follow-up on main (3726128) — nothing for you to do. Welcome aboard!

PR #33, opened 2026-07-23
Media streams: voice/stream sync, application & VoiceOver audio sources

rfiorentino1:
Two additions to the live device-streaming feature from #29, plus a follow-up fix.

## Voice/stream sync (automatic)
Someone singing (voice) over their own streamed instrument (media) arrived ~0.5-1s apart at receivers: the media path carries sender-side latency (operating buffer + the FFmpeg open's standing latency) while voice inserts near-live. Outgoing voice is now delayed by the media path's **measured** latency whenever a live-capture stream runs:
- The loopback connection publishes a media-position → capture-time mapping (`MediaSyncClock`; publisher-elected across FFmpeg's double-opens, invalidated across pad/skip discontinuities).
- `MediaSyncEstimator` measures from our own stream's local playback blocks (hybrid `uSampleIndex`/accumulator position — the index turns out to be unmaintained for local media), snaps a median ~2s in, then tracks drift with corrections applied only on silent chunks so speech never gaps mid-word.
- `VoiceSyncDelayLine` is a deadline FIFO that never reorders: it shrinks by skipping silence, retries when the SDK queue is full, and clears on stream stop so speech returns to live instantly.
- File/URL streams are untouched (the sender monitors those delayed, so they self-align).

Verified: sender-side latency holds ±3 ms over multi-minute streams (5s diagnostics in the audio log), and a second ttaccessible instance stayed in sync for 30+ minutes. A slow residual drift remains on stock clients only — receiver-side media-vs-voice buffering, invisible and unfixable from the sender; happy to discuss separately.

## Application & VoiceOver audio sources
The Stream Audio Device picker (now a plain button popping an NSMenu — an NSPopUpButton with submenus is not VoiceOver-navigable) offers, besides input devices: **VoiceOver's audio**, and **any application's audio** (running apps listed, or any installed app via Select Application…, which works for not-yet-running apps). Tiered by OS:
- macOS 14.2+: Core Audio process taps, with a process-list watcher that re-attaches when the target app quits/relaunches/spawns helpers, and an optional mute-the-source-locally-while-streaming checkbox (never persisted — an accidentally muted VoiceOver would be catastrophic).
- macOS 13.0–14.1: ScreenCaptureKit per-app audio (Screen Recording permission).
- macOS 12: devices only (no public per-app capture API).

System-private `CADefaultDeviceAggregate`/hidden devices are filtered out of device lists.

## Finder/QuickLook fix (second commit)
Selecting Finder streamed silence: QuickLook previews play in helper processes (QuickLookUIService, WebKit.GPU) whose bundle IDs share nothing with com.apple.finder. Tap matching now also accepts processes whose **responsible process** is the selected app — via `responsibility_get_pid_responsible_for_pid` (the attribution TCC itself uses). ⚠️ That's a private-but-ubiquitous libsystem call, resolved via dlsym so its removal would degrade to prefix matching rather than break launch — flagging it explicitly since it's your call whether that's acceptable; without it, Finder/QuickLook (and similar helper-hosted audio) can't be captured.

All strings localized (en + fr). Tests: the device-stream integration test now counts Ogg pages (it still counted AAC frames from before the Opus switch); new VoiceSyncTests cover the estimator, the mapping watermark, and FIFO ordering.


math65 (changes requested):
Thanks Rocco — this is a big, careful PR and the risky core (the voice/stream sync engine) holds up well: I put the deadline FIFO, the `indexTrust` state machine, convergence and ordering through a dedicated correctness pass and nothing survived. The issues below are all peripheral (VoiceOver + CoreAudio lifecycle), not in the sync math.

Grouping: **#1 and #2 are the blockers**; #3–#5 are follow-ups.

---

### 🔴 1. Source picker never announces the current selection to VoiceOver (regression) — `AppDelegate.swift:1602`

The picker is now a plain `NSButton`. `applySelection()` updates `button.title` (line ~2641), but the static `accessibilityLabel` set at line 1602 ("Audio input device") **overrides** the title for VoiceOver on an `NSButton`. So a VoiceOver user always hears *"Audio input device, button"* regardless of the selected source (mic / VoiceOver / Safari…). The previous `NSPopUpButton` exposed the selection as `AXValue`, so this is a regression on the dialog's primary control — for the app's blind-first audience.

**Fix:** in `applySelection()`, also set `button.setAccessibilityValue(spec.displayName)` (keep the label as the field's purpose). This mirrors the existing microphone-button pattern in `ConnectedServerViewController.swift:724-725` (static label + dynamic `setAccessibilityValue`).

### 🟠 2. Process-tap aggregate device created without the device-change suppression the AEC path uses — `ProcessTapCaptureBackend.swift:318`

`buildCaptureLocked` calls `AudioHardwareCreateAggregateDevice` (and on every watcher rebuild, destroy + re-create). The **AEC speaker-tap path does the identical operation** and deliberately guards it: `startSpeakerTapForAEC` calls `extendDeviceChangeSuppressionLocked(2.0)` first, with the comment *"creating the aggregate device triggers kAudioHardwarePropertyDevices which would restart the sound system"* (`+Audio.swift:807`). The streaming path has no such guard, and `ProcessTapCaptureBackend` holds no controller reference to call it.

Caveat, to be fair: `processAudioHardwareChangeLocked` only calls `restartSoundSystem` when `needsAudioReinitializationLocked` returns true (default-device change / chosen device disappearing), so it's not a *guaranteed* restart loop. But with system-default in/out (the common case), if the aggregate perturbs default routing you get a real mid-stream `restartSoundSystem` = mic dropout — and the process-list watcher re-creating the aggregate every ~0.3 s on QuickLook-helper / app-relaunch churn multiplies the exposure. You judged this hazard worth suppressing in the identical AEC case, so it's worth mirroring here.

**Fix:** plumb a suppression hook into `ProcessTapCaptureBackend` (a closure calling `extendDeviceChangeSuppressionLocked`) and invoke it immediately before each aggregate create/destroy, on both initial build and rebuild — mirroring `startSpeakerTapForAEC`.

---

### 🟡 3. HAL process-list listener leaks on every process-source stream stop — `ProcessTapCaptureBackend.swift:59`

`processListListener` is typed as a plain Swift function `(UInt32, UnsafePointer<AudioObjectPropertyAddress>) -> Void`, **not** `AudioObjectPropertyListenerBlock` (`@convention(block)`). Swift re-bridges a plain-function-typed value to a *new* block at each call site, so `AudioObjectAddPropertyListenerBlock` (line ~231) and `AudioObjectRemovePropertyListenerBlock` (line ~247) receive **different block objects**. CoreAudio matches by block identity → the remove never matches, and each stream start/stop permanently leaks one live listener + its retained `controlQueue`. `[weak self]` prevents a crash, but the accumulation is real (each leaked block re-fires on every process-list change). *(Flagged independently by two reviewers and reproduced by compiling a minimal case.)*

**Fix:** `private lazy var processListListener: AudioObjectPropertyListenerBlock = { [weak self] _, _ in self?.scheduleRebuildCheck() }` so a single stable block instance is passed to both add and remove. Optionally check the remove `OSStatus` and log on failure.

### 🟡 4. Accessible label still says "Audio input device" though the control now selects VoiceOver/apps — `AppDelegate.swift:1602`

The PR rewrote `mediaStream.device.prompt.message` to "audio source" (en + fr) but left `mediaStream.device.prompt.deviceLabel` = "Audio input device" / "Périphérique d'entrée audio", which is used as the button's accessible name — inaccurate now that the control also picks VoiceOver and applications.

**Fix:** add a new key `mediaStream.device.prompt.sourceLabel` = "Audio source" / "Source audio" (both `.lproj`) and use it at line 1602.

### 🟡 5. `preselect` can leave `selectedSpec` nil → Stream silently does nothing — `AppDelegate.swift:2634`

Narrow case: macOS 13.0–14.1, no CoreAudio input devices, VoiceOver not SCK-capturable, but application sources exist. `preselect` falls through all branches (device / VoiceOver), leaving `selectedSpec` nil; pressing Stream hits `guard let spec = picker.selectedSpec else { return }` and dismisses the sheet with no stream and no announcement. (On 14.2+ VoiceOver is always preselected, so this can't happen there.)

**Fix:** add a final `applicationSources.first` fallback in `preselect`, or announce when Stream is pressed with no selection.

---

### Checked and deliberately not flagged
- **`VoiceSyncDelayLine` timer not cancelled in `deinit`** — the controller is an app-lifetime singleton, never deallocated; `[weak self]` makes it harmless.
- **0.7 s fallback delay is a dead branch** — true but benign: "voice stays live" is the correct behavior when measurement hasn't converged. Comment-only cleanup.
- **`median()` picks the upper element on even counts** — sub-millisecond bias, far below the 30 ms resnap / 60 ms hysteresis tolerances.

Nice work on the sync engine itself.

rfiorentino1:
All five addressed — plus a correction on the first one that I think is worth your time.

**#1 — the source picker not announcing its selection.** `setAccessibilityValue` on that button does nothing. It cannot: the control is a plain `NSButton`, and the attribute setters can't change an element's **role**, so VoiceOver goes on announcing "button" with no value. The microphone-button pattern doesn't transfer here.

What it needed was the picker machinery from the Mixer app's virtual-accessibility engine — which `MixerVirtualAccessibility.swift` explicitly omitted when it was ported ("the X32 picker/text-edit/section machinery is omitted"). That's now ported: `VirtualPickerConfig`, the `.picker` case (role `.popUpButton`, role description, the selected source as the accessibility value), and `VirtualPickerController` with increment/decrement stepping, so every source can be reached without opening the menu at all. Two deliberate deviations from Mixer: options are read through a closure (this list grows when the user browses for an app), and the behaviour lives in a controller so the **visible** control adopts it — Mixer needs an invisible parallel view tree only because its visible layer is SwiftUI.

**A second bug surfaced while fixing that, and it's worth knowing about generally:** VoiceOver could not enter the applications submenu — it announced a group and dropped focus back to the dialog. The cause was not the menu. **A submenu popped from a control inside an `NSAlert` sheet is not VoiceOver-navigable.** Instrumenting the menu delegate showed the submenu opening and being torn down within ~1 ms whenever a `flagsChanged` event (VoiceOver's own modifiers) was in flight, surviving only when `NSApp.currentEvent` happened to be nil — hence months of it looking intermittent. Presenting the alert app-modally instead of as a sheet fixes it outright, with the menu, its submenu, the items and the control all unchanged. Ruled out along the way, in case it saves you time elsewhere: `autoenablesItems` (items were already enabled), `popUp(… in:)` anchored vs `nil` vs a zero-size anchor view, popping at the mouse location, and `NSPopUpButton` vs `NSButton`. Worth auditing other sheet-hosted menus.

**#2** — suppression hook plumbed from the controller through `AudioDeviceStreamSource` and invoked before every aggregate create *and* destroy, on initial build and on every watcher rebuild.

**#3** — `processListListener` is now typed `AudioObjectPropertyListenerBlock`; the remove status is logged on failure.

**#4** — I renamed the key to `mediaStream.device.prompt.sourceLabel` rather than adding a second one, since `deviceLabel` had no other use. Say the word if you'd rather keep both.

**#5** — `preselect` falls back to the first application source.

**Cross-PR warning:** this branch and #36/#37 each introduce their own `channelPathLocked(instance:channelID:)` in different files. They don't conflict in git, but whichever merges second will need a one-line de-duplication.

math65:
Reviewed the whole branch merged into main rather than in isolation, and pushed the follow-ups here since #35/#36/#37 are all merged now.

First, the two findings you flagged in your last comment are both real and both bigger than this PR. The submenu-in-an-NSAlert-sheet one especially — a submenu torn down within a millisecond whenever a `flagsChanged` is in flight, which is to say whenever a VoiceOver user is the one trying to open it, explains a class of "intermittent" behaviour I would never have found by reading. Leaving the full reasoning in a comment above the app-modal presentation was the right call; without it someone converts that back to a sheet in a year. And you were right that `setAccessibilityValue` on an `NSButton` cannot work — the setters do not change an element role.

**Rebased onto main** (24 commits, base 1.8.0). Three conflicts, all mechanical: six blocks in `AppPreferences` where both sides add fields (your `deviceStreamLastSource` / `deviceStreamVoiceSyncTrimMSec` against 1.9.0's `languagePreference` / `hasChosenInitialLanguage`), the controller init, and `DeviceStreamSourceTests`. Note your cross-PR warning did not apply here — this branch never had a `channelPathLocked`, so there was nothing to dedupe; that was only #36 and #37.

**Fixed: stopping a stream ran the entire backend teardown on the SDK's serial queue.** `finalizeMediaStreamingLocked`, `destroyLocked` and the three start-failure paths all called `AudioDeviceStreamSource.stop()` from it. `SCKAudioCaptureBackend.stop()` waits on a semaphore for up to 2 s, and `ProcessTapCaptureBackend.stop()` takes `controlQueue`, which may be mid-rebuild — the 0.1 s settle plus aggregate create/destroy. The process-list watcher rebuilds when the target app quits, which is precisely when someone stops the stream, so the bad case is the likely one. On the SDK queue that stalls the message loop for everyone, and `disconnectSynchronously()` puts the main thread behind that same queue. Added `stopAsynchronously()`: closes the loopback server synchronously (cheap), hands backend teardown to a global queue, inverted order so the ring has no reader while the backend winds down. Worth contrasting with your start path, which explicitly dispatches off the queue with a comment saying why — the same care just had not reached the stop side.

**Fixed: VoiceOver announcement priority was being dropped in `MixerVirtualAccessibility`.** `NSAccessibility.NotificationUserInfoKey.priority` expects an NSNumber; passing the `NSAccessibilityPriorityLevel` enum itself bridges to an opaque box, so AppKit cannot read the level and the announcement silently loses high priority — it becomes interruptible by whatever VoiceOver is already saying. Every other announcement in the app passes `.rawValue`; this file was the only one that did not. Three of the four sites predate this branch (mixer slider and toggle announcements, shipped since 1.7.0), so this is a live bug rather than something you introduced — fixed all four together.

Two smaller things I did **not** change, both yours to judge:

- `SCKAudioCaptureBackend`'s `resampleScratch` is sized at 32768 frames where the process tap computes a worst case for an 8 kHz source. Harmless today since SCK is configured for 48 kHz and the resample path is short-circuited, but the asymmetry between two otherwise-twin backends is a trap if that config ever changes.
- `PCMRing.write`'s comment still says "only ever called from the single AUHAL callback thread". There are three possible producers now. Not a bug — they are never concurrent, and `AudioDeviceStop` fences the handover — but the comment no longer describes the code.

I also want to correct something, because I nearly sent you a bogus finding: I had flagged `NSWorkspace.runningApplications` being read off the main thread in `matchedProcesses()`. The AppKit header is explicit that it is thread safe and returns atomically, and `NSRunningApplication` is `NS_SWIFT_SENDABLE`. Your code is right and mine would have been the regression. The only real caveat is that background reads only refresh on the main run loop's next turn, which your 0.3 s debounce and re-triggering listener already absorb.

Rest of what I read and found sound: the SPSC ring (release/acquire, no RT allocation), the `AudioBufferList` allocation in `DeviceInputCaptureBackend` avoiding the single-buffer sizing trap this codebase has been bitten by, tap scratch buffers allocated once for the backend's life and never on rebuild, the delay line confined to the controller queue, and `.measureOnly` skipping the PCM copy when the user is not monitoring. Also: mute-while-streaming being deliberately non-persisted and off by default, on the grounds that an accidentally-muted VoiceOver would be catastrophic — exactly the right instinct.

math65:
Follow-up after merging: I tested this on a live server tonight and hit a blocker, so three commits went straight to main (`0daf796`, `22703b8`, `6caf7bc`). Details here because the root cause is worth knowing, and because it came from the fix you made this morning.

**No audio source could be selected at all.** Opening the picker worked, the menu highlighted and dismissed normally — and the selection was silently discarded. The control kept showing the preselected device whatever you picked, VoiceOver or a plain input device alike.

The cause is `e28754d`, the commit that moved the dialog from `beginSheetModal` to `runModal` so the applications submenu would be VoiceOver-navigable. **In a modal session AppKit never delivers an `NSMenuItem`'s action.** Your `autoenablesItems = false` handled the other half of the same modal behaviour — items rendering disabled — but nothing was dispatching the action. So the submenu became reachable and simultaneously useless.

How I established it, since it is not obvious from reading:

- Reproduced on macOS 26.5, **Debug and Release alike**, so not an optimiser/lifetime issue — that was my first hypothesis and it was wrong.
- **Mouse and keyboard alike**, so not a hit-testing problem.
- Instrumented both ends: the menu logged its `popUp` (6 items, `modal=1`), the action never fired once.
- Retargeting the items at the button *inside* the modal window did not help either — a modal session drops the action regardless of target.

**The fix keeps your submenu.** The dialog is no longer an `NSAlert` at all: it is a sheet hosting a plain view controller (`MediaStreamSourceViewController`), and the source button keeps exactly the menu you built — devices and VoiceOver at the top level, applications in a submenu, browse entry at the end. Outside a modal session the menu, the submenu and their actions all behave normally. And the app-modal presentation is no longer needed, since it only ever existed to work around a submenu inside a sheet-hosted `NSAlert` — a plain sheet is neither.

Verified end to end: menu and submenu open, choosing an application from the submenu applies, the mute checkbox enables and greys with the source type, the choice is announced, Return confirms from the button, streaming starts (`process tap: capturing 1 process(es) for VoiceOver: com.apple.VoiceOver`) and stops cleanly.

Two things found along the way:

- **Sheets could stack.** Source discovery is asynchronous, so the "already streaming" guard was evaluated at keystroke time rather than at presentation — two quick ⌘⌥A presses each finished discovery and each presented a sheet. I reproduced it by accident. Now guarded on both an existing sheet and an active stream.
- **The menu never advertised the feature.** The entry and the dialog title still read "Stream Audio Device" after this PR added VoiceOver and per-application sources — which is how Mathieu came to ask where application streaming had gone. Both now read "Stream a Device or Application".

`DeviceStreamSourcePicker` and the `DeviceStreamSourceButton` alias are gone, having no callers left. **`VirtualPickerConfig` / `VirtualPickerController` are untouched in `MixerVirtualAccessibility`** — they have no consumer now, since a sheet-hosted menu needs none of it, but that is where the Channel Mixer would want them given its visible layer is SwiftUI. Your call whether they stay.

None of this takes away from the diagnosis in your last comment: the submenu really was being torn down inside a sheet-hosted `NSAlert`, and the `flagsChanged` instrumentation that found it is the kind of thing nobody finds by reading code. The app-modal presentation just traded one failure for a worse one, and only a live test surfaced it — which is on me as much as anyone, since I merged it.

PR #34, opened 2026-07-23
Proposal: drop the App Sandbox so global hotkeys can truly swallow keys

rfiorentino1:
**Still a draft, still your call — this reverses a decision that was deliberately yours (a63aac7).** The three things you asked to settle before taking it are now addressed (last commit); the rest of this is why I think the problem is sharper than the ⌘⇧A example I opened with.

## The problem, stated properly

My original framing — ⌘⇧A leaking into Safari — is the *opt-in* case, and it undersold this. Two things are true today:

- **`pushToTalkGlobal` defaults to `true`**, and push-to-talk has no default binding. So the first time a user records a push-to-talk key — an ordinary letter, most likely — it is global by default. A PTT key is **held for the whole transmission**, so that key and its auto-repeat go into whatever app is frontmost for as long as they talk.
- **The mic toggle is opt-in** (`muteHotkeyGlobal` defaults to `false`), but opting into "works when another app is in front" is not opting into "also triggers Safari's autofill". Both are the same defect; one is just reached sooner.

The bar I think we should hold is simply: **a key command the app defines should not reach any other app.** Sandboxed, we cannot meet it.

## There is no sandboxed fix

All three mechanisms have been tried and are dead ends:

| Mechanism | Result |
|---|---|
| Carbon `RegisterEventHotKey` | Does **not** swallow — the frontmost app still receives the key. |
| Listen-only `CGEventTap` (Input Monitoring) | Detects only, by construction. Current behaviour. |
| Consuming `.defaultTap` / NSEvent global monitor | Would swallow, but requires **Accessibility**, which the App Sandbox blocks. |

The modifier-chord recommendation covers toggles but not push-to-talk: a chord held for the duration of speech modifies every keystroke made while transmitting, which is worse for VoiceOver users, whose navigation is modifier-based. F13–F19 are genuinely safe and genuinely absent from laptop and Magic keyboards.

## Addressed since your review

1. **Accessibility prompt fires once, ever** — not once per launch, and skipped entirely while sandboxed. Preferences carries a standing explanation of what Accessibility buys, where to grant it, and that until it is granted the key also reaches the frontmost app. (The old help text asserted the opposite; this branch made it false.)
2. **Security-scoped bookmarks** — this turned out to be a live bug rather than a question. Outside the sandbox `bookmarkData(.withSecurityScope)` throws and `startAccessingSecurityScopedResource()` returns `false` for an ordinary URL, meaning "no scope needed" rather than "denied". Both were treated as failures, so the recording folder was never remembered and was **re-prompted on every recording**. `AppSandboxState` now centralizes the decision; recording folder, sound-file store and `.tt` config import all branch on the real sandbox state.
3. **TCC on Documents/Desktop** — added `NSDocumentsFolderUsageDescription`, `NSDesktopFolderUsageDescription` and `NSDownloadsFolderUsageDescription`, so the prompt explains itself instead of appearing bare.

## Migration, verified live

One-time, never-clobber import at first unsandboxed launch: every preference domain replayed **through** UserDefaults (never by copying plists into `~/Library/Preferences` — cfprefsd ignores files written behind its back and flushes over them; we hit exactly that), plus the Application Support payload. Keychain items carry over untouched — their ACLs bind to the code signature, not the sandbox.

Tested on a real long-lived profile: saved servers, preferences, profiles and keychain passwords all intact, and ⌘⇧A fully swallowed with Safari frontmost. Users updating via Sparkle keep one signature throughout, so the only new user-visible step is granting Accessibility once — release-notes material.

## If you'd rather not

Then I think the honest alternative is to stop offering global bindings on ordinary keys at all, rather than shipping a mode where our key commands land in other people's apps. That is a real reduction in the feature, which is why I'd rather have this conversation than quietly pick either one.


math65:
Thanks for making this a discussable draft rather than a fait accompli — and for the quality of the prototype: the migration replays preferences through the UserDefaults API instead of copying plists behind cfprefsd's back (the v2 marker tells me you hit that trap for real), the container is left in place as a safety net, and the tested-on-a-real-profile note is appreciated.

I'm not against dropping the sandbox on principle. But if we can avoid it, I'd prefer to, for two reasons:

1. **Defense in depth.** We run libTeamTalk5, a large closed-source C/C++ library that parses untrusted network input (server packets, other clients). Today a memory-corruption bug in the SDK is confined to the container; unsandboxed, it's the whole user account. That's exactly the app profile the sandbox was designed for, and it protects *every* user — while true key-swallowing benefits the subset who use a *global* hotkey bound to a *regular* key.
2. **It's a one-way door in practice.** Coming back later means a reverse migration — fresh data in ~/Library would be invisible to a re-sandboxed build.

What makes me want to wait: #31 already gives us a clean escape hatch. Bindings are configurable now, and a pure-modifier chord (say ⌘⌃) doesn't leak *anything* into the frontmost app — modifiers alone type nothing. As you noted, even the unsandboxed consuming tap can't swallow flagsChanged, so for modifier chords the two builds behave identically anyway. F13–F19 are similarly harmless. So my proposal: let's ship 1.9 with #31 as is, recommend modifier-only chords / F-keys as the global binding in the release notes and in the Preferences help text, and see what beta feedback says. If the non-swallowing pain is still real after that, this PR is the serious, tested path and I'll take it — with three things to settle in review first: the Accessibility prompt firing on every launch (should be once + a hint in Preferences), what happens to existing security-scoped bookmarks for the recording folder once unsandboxed, and the new TCC prompts on Documents/Desktop.

Leaving this open as a draft in the meantime. Thanks again — the hotkey engine itself is great work, and it lost nothing by staying sandboxed.

rfiorentino1:
Your three conditions are addressed in the latest commit, and I've rewritten the description because my original framing undersold the problem.

The ⌘⇧A/Safari example I opened with is the *opt-in* case. The one that forces the issue is that **`pushToTalkGlobal` defaults to `true`** and push-to-talk has no default binding — so the first time a user records a PTT key, it is global by default, and a PTT key is **held for the whole transmission**. That key plus its auto-repeat goes into whatever app is frontmost for as long as they talk. The modifier-chord recommendation doesn't cover it: a chord held for the duration of speech modifies every keystroke made while transmitting, which is worse for VoiceOver users whose navigation is modifier-based, and F13–F19 are absent from laptop and Magic keyboards.

The conditions:

1. **Accessibility prompt fires once, ever** — persisted, and skipped entirely while sandboxed where the grant is unusable. Preferences carries a standing explanation instead: what Accessibility buys, where to grant it, and that until it is granted the key also reaches the frontmost app. The old help text asserted the opposite, which this branch made false.
2. **Security-scoped bookmarks** — this turned out to be a live bug rather than an open question. Outside the sandbox `bookmarkData(.withSecurityScope)` throws and `startAccessingSecurityScopedResource()` returns `false` for an ordinary URL, meaning "no scope needed" rather than "denied". Both were being treated as failures, so the recording folder was never remembered and was re-prompted on *every* recording. `AppSandboxState` centralizes it; recording folder, sound-file store and `.tt` config import all branch on the real sandbox state.
3. **TCC on Documents/Desktop** — added the three usage-description strings so the prompt explains itself rather than appearing bare.

Still a draft and still your call. If you'd rather keep the sandbox, I think the honest alternative is to stop offering global bindings on ordinary keys altogether, rather than ship a mode where the app's own key commands land in other people's apps — that's a real reduction in the feature, which is why I'd rather settle it with you than quietly pick one.

rfiorentino1:
Bringing this back, and marking it ready rather than draft. Rebased onto current `main` (1.10.0) — the two #31 commits dropped out with the merge, so what's left is the three sandbox commits plus the review-conditions one.

I've been running the unsandboxed build as my daily driver since the 22nd — four days, real servers, real sessions. Migration v2 restored every saved server and channel password (the keychain needed nothing; its ACLs bind to the code signature), and my global mute chord is genuinely swallowed: it fires the mic toggle and Safari never sees it. That's the field evidence this PR didn't have when you deferred it.

On the escape hatch: I don't think it reaches the users who need it. The shipped default is ⌘⇧A — a regular character key — so out of the box every user is in exactly the leaking case the release notes would be telling them to avoid. F13–F19 are clean, but they don't exist on a laptop keyboard or a Magic Keyboard, which is most Macs. And modifier-only chords are fine for a *toggle*, but they're the wrong tool for push-to-talk specifically: the chord is held for the entire transmission, so everything typed while talking arrives modified — and for a VoiceOver user, whose navigation is *entirely* modifier+key, holding ⌘⌃ to talk changes what every single key does while they're talking. That one isn't fixable by dropping the sandbox either (flagsChanged can't be suppressed at all), which is why I don't think "recommend chords" is a substitute for making regular-key bindings work.

On the one-way door: the migration isn't a one-off script, it's a generic replay of keys through the UserDefaults API. Running it in the other direction is the same code with source and destination swapped, and the container is left in place untouched rather than deleted. If you ever want the sandbox back, the reverse path is a small amount of work against a mechanism that now has field miles on it.

Your three conditions are addressed in the last commit: the Accessibility prompt fires once per launch and only when a global binding is actually in use, with a hint in Preferences; the recording-folder bookmark and the Documents/Desktop TCC strings are covered there too.

The defense-in-depth argument I don't dispute — libTeamTalk5 is exactly the kind of code the sandbox exists for. I just don't think the global hotkey is a niche feature we can route around: it's the feature, for the people this app is for.

math65:
Before this goes any further I want to put a measurement in front of you, because it bears directly on the premise of this PR — and I'd rather you shot it down than have me act on it.

`HotkeyMonitor.swift` says Carbon was tried and rejected:

> (Carbon RegisterEventHotKey was tried and rejected — it doesn't swallow either, and can't do bare keys / modifier-only chords.)

I measured it here and got the opposite result on both counts. **But I don't think you made that up**, so the interesting question is what differs between our two machines — not which of us is right.

### What I ran

A minimal `.app` signed with `com.apple.security.app-sandbox`, `.accessory` policy so it is never frontmost. Sandbox proven active at runtime rather than assumed: writing outside the container fails and `NSHomeDirectory()` returns `~/Library/Containers/<id>/Data`. It registers ⌥J and logs `kEventHotKeyPressed` / `kEventHotKeyReleased`. ⌥J because it inserts a character when it is *not* swallowed, so the front app's document answers the question by itself.

With TextEdit in front, keys pressed **physically** on the keyboard, **VoiceOver running**:

| Step | Result |
|---|---|
| ⌥J, hotkey registered | handler fired — nothing inserted, anywhere |
| Control: plain "A" | inserted normally |
| ⌥J after unregistering | "Ï" inserted |

Press/release, uptime seconds:

| Gesture | PRESSED | RELEASED | Held |
|---|---|---|---|
| short tap | 288817.303 | 288817.436 | 133 ms |
| held ~2 s | 288821.562 | 288823.391 | 1829 ms |

One `PRESSED` per press even when held — no autorepeat storm, unlike the tap, which has to filter `keyboardEventAutorepeat` by hand.

And `RegisterEventHotKey` returned `0` for F13 bare, a bare letter, ⌘⇧A and ⌥⇧⌘Space. The only form it genuinely cannot express is the modifier-only chord: it takes a key code, and 0 is a real key.

### So: what did you see?

If that holds on your machine too, this PR's premise mostly dissolves — hotkeys already swallow inside the sandbox, and dropping it would only buy the modifier-only chord. If it does *not* hold on your machine, that is much more interesting than my result, and I'd rather know before touching anything.

One difference I expected to matter turned out not to: **VoiceOver was running throughout**, and the physical keypresses above were made with it active. That was the configuration most likely to break this — VoiceOver takes a lot of keystrokes — and it changed nothing. So that one is already ruled out rather than pending.

Ways our setups could still differ:

- **Secure input.** A password manager or a focused secure field neutralises Carbon hotkeys system-wide, and it can stay latched long after the field is gone. `ioreg -l -w 0 | grep kCGSSessionSecureInputPID` shows whether something is holding it.
- **macOS version.** I'm on 26.
- **Which chord you tested.** A chord already owned by the system or another app would look like "doesn't swallow" for an unrelated reason.
- **Signing.** Mine was signed with a real Apple Development identity; ad-hoc or unsigned can behave differently under TCC.

Happy to send the test app, or you can rebuild it in about twenty lines — say which is easier.

### What I have not measured

- Behaviour under secure input, beyond knowing it neutralises Carbon hotkeys.
- Whether a Carbon hotkey and the SwiftUI menu key equivalent can hold the same chord without one shadowing the other. If Carbon swallows first, the menu item would keep advertising a shortcut it no longer answers.
- Migration of bindings already stored in preferences.

Tracking this in #41 so it doesn't get lost in this PR's thread.

rfiorentino1:
Closing this. I reproduced Mathieu's Carbon measurement on my machine (#41) — `RegisterEventHotKey` swallows the chord from inside the sandbox, with no permission at all. That removes the premise this PR was built on.

What's left over is the modifier-only chord, which Carbon can't express — but nothing in here fixes that either, since we never consume `flagsChanged`. So there's no version of this PR that buys anything today.

The branch stays up if it's ever needed: the preferences migration is a generic replay through `UserDefaults` and runs in either direction. Reopening is cheap if the modifier-chord case ever forces it.

PR #35, opened 2026-07-23
Move all users in a channel at once

rfiorentino1:
## What

Adds an accessible way to move **multiple** users to another channel in one operation — today an admin can only move users one at a time.

A new action **"Move all in channel…"** appears in the server-tree context menu and as a VoiceOver custom action on both channel and user rows. It opens a checklist sheet:

- A **VoiceOver-navigable list** of the channel's occupants (labeled "Users"), each a checkbox — all pre-checked, so the common "clear the room" case is just open → pick destination → Move. Deselect anyone to leave behind.
- **Select All / Deselect All** buttons and a **target-channel** picker.

It resolves the target channel from **either** a selected channel row **or** the channel of a selected user, so it lights up wherever you invoke it (right-click a person → move everyone in *their* channel).

## Notes

- Reuses the existing per-user `TT_DoMoveUser` path (`moveUser(...)`) — no new SDK plumbing.
- **Permissions:** admin/operator-gated, and only enabled when the channel has occupants — mirroring the existing single-user move rule. The single-user "Move to a channel…" is untouched.
- New user-facing strings added to both `en.lproj` and `fr.lproj` (French in vouvoiement).
- Accessibility-first: the whole flow works from the keyboard/VoiceOver without needing mouse multi-selection.


math65:
Thanks — this is the cleanest of the three PRs I reviewed today: strings are correctly localized (vouvoiement is fine), the serial-queue rule is respected, sheet retention is correct, and there are no blocking bugs. There's one real accessibility bug and a set of VoiceOver-polish items I'd want addressed before merge, since the accessible-first flow is the whole point of the feature.

## Correctness

**1. The VoiceOver custom action moves the wrong channel** (`+OutlineDelegate.swift`). The action is attached to a specific channel row — the closure has `channel` in scope and even uses it for its `!channel.users.isEmpty` guard — but it calls `moveChannelUsersAction(nil)`, which re-resolves the target from `outlineView.selectedRow` via `bulkMoveTargetChannel()`. Moving the VoiceOver cursor doesn't change the selected row, so a VO user browsing to channel B and invoking "Move everyone here" while row A is selected moves **channel A's** occupants — or silently no-ops if nothing is selected. The codebase already has the selection-independent idiom for exactly this: `performDefaultAction(for: node)` (`+ChannelActions.swift:474`, with the comment "la ligne sous le curseur VoiceOver, sans dépendre de la sélection"). Capture `channel.id` in the closure the same way.

**2. Bulk-move failures stack N modal alerts, or vanish** (`+UserActions.swift`, the `onMove` loop). The permission gate is target-blind: `canModerate` checks `isChannelOperator` of the *local user's own* channel, not the channel being emptied — so the action can fire on a channel the server will reject. On failure, each user calls `presentError` (no local override → AppKit's modal `NSResponder.presentError`), so N synchronous failures stack N blocking dialogs (worst under VoiceOver). And async rights-rejections come back as `>0` → `.success`, so they're dropped silently — the room stays put with no message. I'd aggregate into one "Moved N of M" result and, ideally, check operator rights on the target channel before enabling the action.

## VoiceOver polish (the important part for this app)

- **Select All / Deselect All give no VO feedback** — they flip `checkbox.state` with no `.announcementRequested`, so a blind admin hears nothing and must arrow through every checkbox to confirm. `confirm()` already posts an announcement for the empty case; an "All selected" / "All deselected" post is the consistent fix.
- **The occupant list isn't exposed as a list** (no group role / "item 3 of 12" context / boundary), and **the header has no `.isHeader` trait** — so rotor heading-nav finds nothing and the checkbox run has no position/count. (Individual checkbox state is fine — native `NSButton`.) CLAUDE.md: "All section headings use `.accessibilityAddTraits(.isHeader)`."
- **The target-channel popup shows bare channel names** — TeamTalk allows same-named subchannels under different parents, so multiple "General" entries are indistinguishable to VO. `ConnectedServerChannel` already carries `pathComponents`; use `$0.pathComponents.joined(separator: " / ")` as the title. (Pre-existing — the single-user `performMove` picker at `+UserActions.swift:385` does the same — but worth fixing in both.)
- **No result announcement** after the move. Pre-existing pattern (single-user has the same gap), but a bulk action especially warrants an aggregate "Moved N users to <channel>".

## Layout / structure (smaller)

- **Long occupant lists open scrolled to the bottom** (`MoveUsersViewController.swift`). The `documentView` (`listStack`) is pinned only leading+top to the clip view — no bottom/width tie — and the clip view isn't flipped, so a list taller than the sheet shows the *last* users first until you scroll up. Pin bottom + width to the clip view (or use a flipped document view).
- **Re-implements the existing `performMove`** (`+UserActions.swift:355`) — flatten channels, pick target, loop `moveUser` — with divergent exclusion rules (`commonChannelID` vs `excludingChannelID`). Worth noting the new path actually *fixes* a latent `performMove` bug (it offers the source channel as a destination for multi-user moves). I'd extract a shared "build target list + loop `moveUser`" core taking an explicit excluded-channel-id, and layer the checklist UI on top (the subset-deselect UX is legitimately new).
- **Frozen occupant snapshot** — a user who leaves between opening the sheet and pressing Move becomes a failed move (feeds finding #2). Minor; a re-validation on confirm would avoid the doomed calls.
- **`preselected: Set<Int32>` is dead generality** — the only caller always passes the full candidate set, so no checkbox ever starts unchecked. Drop the parameter and set every box to `.on`.

Solid groundwork — it mostly needs the VO action fixed and the accessibility affordances (announcements, list/heading roles, path-qualified channel names) finished, which is the highest-value part for this app. Happy to pair on the `performMove` unification.

rfiorentino1:
All addressed. Three notes where I went further than, or short of, the review.

**Correctness**

- **VoiceOver action target** — the action now carries its own channel ID and re-resolves it from the live session, via an explicit-channel entry point the menu item shares. No dependence on `selectedRow`.
- **Permission gate** — you were right that `canModerate` is target-blind, and it can't be fixed at the model level: `ConnectedServerUser.isChannelOperator` is computed as `TT_IsChannelOperator(…, user.nChannelID)`, i.e. the user's *own* channel. Added `canMoveUsers(fromChannelID:)`, which asks the SDK for the account-wide move right or operator status on that specific channel; both the menu validation and the VoiceOver action use it.
- **Stacked alerts** — moves now report one aggregate "Moved N of M users to …", announced for VoiceOver plus at most one alert when some failed.

**Short of the review, deliberately:** async rights-rejections still aren't correlated back to individual moves. Doing that needs command-ID tracking the controller doesn't have, and it would affect every admin command, not just this one — it felt wrong to smuggle that into a bulk-move PR. Happy to do it separately if you want it.

**VoiceOver** — Select All/Deselect All announce; the occupant list carries the list role and per-item position ("Alice, 3 of 12"); the sheet header carries `AXHeading` (by raw value, since the typed constant is macOS 26+ and this app targets 12); destination channels are listed by path.

**Structure** — extracted `moveDestinationChannels(excluding:)`, `commonChannelID(of:)` and `commitMove(userIDs:to:presentingWindow:)`, shared with `performMove`. `commitMove` drops users who left while the picker was open. `preselected` is gone.

**Worth calling out — this changes shipped behaviour** in the existing single/multi-user move dialog, since it now shares that core:

- channels are listed by path rather than bare name
- the move announces its result, where it was previously silent
- multiple failures give one aggregate alert instead of one per user
- the source channel is now excluded for multi-user moves too (the latent bug you spotted)

All improvements as far as I can tell, but they are visible to existing users, so flagging rather than burying them.

math65:
Everything I asked for is addressed, and the parts I re-checked hold up: `moveUser` completes on the main queue in both of its branches, so `commitMove`'s counters and the `DispatchGroup` are sound; `displayPath` correctly drops the root element; and the VoiceOver action does now carry its own channel ID. Flagging the shipped-behaviour changes rather than burying them was the right instinct — I am happy with all four.

I pushed two follow-ups straight to the branch rather than send you another round trip.

**1. The permission check ran a `queue.sync` per outline cell.** `canMoveUsers(from:)` is called from `outlineView(_:viewFor:)`, and went down into `canMoveUsers(fromChannelID:)`, which hops synchronously onto the TeamTalk queue. That is one blocking hop per populated channel row, on the main thread, on every tree reload — against the queue the message loop occupies continuously and that `connectAndLoginLocked` can hold for ~10 s. The `isAdministrator` shortcut only spares admins, so ordinary operators paid it every time. It composes badly with #36 specifically: a tree rebuild during a reconnect attempt would freeze the main thread for the length of the attempt, and a reconnect is precisely when the tree gets rebuilt.

The right now rides on `ConnectedServerChannel.canMoveUsersOut`, computed in the session snapshot on the queue that already owns the SDK — account-wide right resolved once per snapshot, operator status once per channel. The view reads the snapshot and nothing else, which left `canMoveUsers(fromChannelID:)` with no callers, so it is gone. Your diagnosis of *why* `isChannelOperator` could not answer the question was right; it just needed to be answered somewhere other than the render path.

**2. Two VoiceOver strings had no singular.** A partial move of one user read "1 utilisateurs sur 3 déplacés", and Select All in a one-occupant channel read "Les 1 utilisateurs sont sélectionnés" — the one-occupant case being fairly common for a bulk move. Added `.one` variants following the existing `connectedUsers.count.one` convention. Both forms take the same arguments on purpose, so no positional specifier is skipped.

**Separately, and worth more than either fix:** this branch was based on **v1.6.0, 108 commits behind main**. GitHub called it mergeable because nothing conflicts textually, but it had never seen 1.7.0, 1.8.0 or 1.9.0 — so the app you built and tested against has no Channel Mixer and no `OutputAudioRenderEngine` (#22), and the `ttaccessibleTests` target does not exist on that base at all. I merged main in: it builds and the test suite passes, so nothing drifted at the API level — but your manual testing did not cover current code. #36 and #37 are on the same base; #33 is on 1.8.0.

PR #36, opened 2026-07-23
Reliably auto-reconnect and rejoin the same channel after a drop

rfiorentino1:
## Problem

When a server drops (restart, network blip, ping timeout), the client showed a "connection lost" dialog and booted you to the server list — it never actually reconnected, even with the *Auto-reconnect* preference on. The auto-reconnect machinery existed but never worked, due to a few ordering bugs where state was wiped right before it was needed.

## Fixes

- **`startReconnectTimerLocked()` called `cancelReconnectLocked()`**, which nulled the `reconnectRecord`/`reconnectPassword` the caller had just set — so the first retry bailed at its guard and booted the user instead of retrying. Now it cancels only the timer.
- **Only the idle `CON_LOST` path honored the preference.** A server-restart logout (`MYSELF_LOGGEDOUT`) and drops caught in the command-completion loop booted unconditionally. All three now route through one `handleServerDropLocked()` that honors `autoReconnect`. (Safe: the app never calls `TT_DoLogout`, and a kick/ban is `MYSELF_KICKED`, handled elsewhere — so `MYSELF_LOGGEDOUT` can only mean the server dropped us.)
- **`attemptReconnectLocked()` read the rejoin target *after* `cancelReconnectLocked()` had cleared it**, so every reconnect landed in the root channel. The target is now snapshotted before cancelling.

## Rejoin by path

Rejoin now keys on the channel **path** (resolved via `TT_GetChannelIDFromPath`) instead of the numeric channel ID, so it survives a full server restart that reassigns IDs — and preserves the channel password across the drop. Retries every 5s until the server returns, then rejoins the exact channel you were in.

Adds reconnect diagnostics to the audio log (`Connection drop …`, `Reconnect attempt …`, `Reconnect: rejoining channel by path …`).

## Testing

Verified live against a real server restart: `CON_LOST` → retry loop → `rejoining channel by path /Lounge/` → reconnected into the same channel, no dialog.


math65:
Thanks for this — the core mechanism is right, and the three ordering bugs you identified in the commit message are real fixes. I ran a deep review and found a set of issues I'd want resolved before merge. Most of them trace back to one root cause, so I've grouped them that way.

## Root cause: reconnect-state lifecycle + `cancelReconnectLocked`'s dual role

`cancelReconnectLocked()` does two things at once — stop the timer **and** wipe the rejoin state — and that coupling is behind four of the findings. The PR works around it twice (the inlined 3-line timer teardown in `startReconnectTimerLocked`, and the five snapshot locals in `attemptReconnectLocked`) instead of splitting it. I think the deeper fix is:

1. Extract `stopReconnectTimerLocked()` (timer only), called by both `startReconnectTimerLocked` and `cancelReconnectLocked`.
2. Move the `cancelReconnectLocked()` call in the success path to **after** the rejoin. I verified this is behavior-preserving on the serial queue (nothing between the current call site and the rejoin reads or needs the reconnect state cleared, and the fire-and-forget `TT_Do*` rejoin can't re-enter `handleServerDropLocked`). That removes all five snapshots.

Doing that also fixes these two directly:

- **Second drop never reconnects** (`+Connection.swift:143`). After one successful reconnect, `reconnectPassword` is nil'd by `cancelReconnectLocked` and never restored (its only setter is `connect()`). So `handleServerDropLocked`'s `guard let password` fails on the *next* drop and the user is booted despite `autoReconnect` — the feature works only once per session.
- **Stuck on "reconnecting…" forever** (`+Connection.swift:215`). If `ensureOutputAudioReadyLocked()` throws on the success path (output device gone during the outage), the timer + credentials are already wiped by the early cancel, and the catch publishes nothing — the UI never leaves the reconnecting state. Moving the cancel after the rejoin closes this too.

## Correctness — needs a fix

- **A server kick/ban triggers the reconnect loop** (`+Connection.swift:577`). The commit note says "a kick/ban is MYSELF_KICKED, so reconnecting on MYSELF_LOGGEDOUT is safe" — but the `MYSELF_KICKED` handlers (`:641`, `:1087`) only append history and set no flag, so the `MYSELF_LOGGEDOUT` the server sends *after* a server-kick routes into `handleServerDropLocked`. The reference client agrees this is unsafe: `mainwindow.cpp:1734` deliberately does **not** arm its reconnect timer on `LOGGEDOUT` (only `CON_LOST` does, `:1078`). Result: a kicked user silently rejoins in 5s; a banned user hammers the server every 5s forever. I'd suggest an explicit "expected logout / just kicked" flag rather than relying on the comment-level assumption.
- **An armed reconnect clobbers a manual connect to another server** (`+Connection.swift:30`). `connect()` never calls `cancelReconnectLocked()`. If server A drops and the user then connects to server B, the A-timer fires 5s later: on failure its catch runs `destroyLocked()` which tears down B's *live* `self.instance`; on success it overwrites B's session with A (ghost user on B, leaked instance). `connect()` needs to cancel any armed reconnect.
- **A full SDK instance leaks per failed attempt** (`+Connection.swift:241`). In `attemptReconnectLocked`, `self.instance` is only assigned on success, so on failure `destroyLocked()`'s `if let instance` teardown is a no-op — the local instance from `createInstanceLocked()` is orphaned with its threads/socket. Every 5s tick during an outage leaks another (and each tick after the first is a cold `TT_InitTeamTalkPoll`, since the warm `reusableInstance` is consumed and never refilled). The catch should `TT_Disconnect`/close the local instance (or restash it in `reusableInstance`).
- **No backoff, no cap, and the retry blocks the serial queue** (`+Connection.swift:185`). The 5s repeating timer runs on the serial queue where `connectAndLoginLocked` can block ~10s against a black-holed host, so attempts run back-to-back with no gap and no give-up path. `disconnectSynchronously()` (`queue.sync` from the main thread) can then freeze the UI up to 10s. I'd go with a one-shot timer re-armed at the end of each *failed* attempt, plus exponential backoff (5s→10s→30s→60s) and eventually surfacing "connection lost".

## Correctness — smaller / edge cases

- **`MYSELF_LOGGEDOUT` captures an empty channel** (`+Connection.swift:144`). The SDK resets `m_mychannel` *before* posting the event (`ClientNode::HandleLoggedOut`), so `TT_GetMyChannelID` returns 0 and the rejoin has nothing to key on. Less urgent since a hard restart surfaces as `CON_LOST` (channel intact — matches your live test), but the LOGGEDOUT path won't rejoin.
- **Numeric-ID fallback can join the wrong channel** (`+Connection.swift:160`). When the path fails to resolve after an ID-reassigning restart, joining by raw `lastID` with no name/path check can land the user in an unrelated channel. This is exactly the case the path-based approach was meant to avoid — I'd rather return false and let `autoJoinAfterLoginLocked` handle it than join by stale ID. Also note the asymmetry: the path branch writes `channelPasswords[channelID] = password` (`:150`), the ID branch doesn't.
- **Command-path drop double-signals the UI** (`+Connection.swift:1011`). `handleServerDropLocked` publishes "reconnecting", then the code still `throw`s `connectionFailed`, which unwinds into ~15 command callers and pops a modal `.critical` NSAlert + VoiceOver announcement on top of the reconnecting state. Confusing, especially with VoiceOver.

**Two findings I deliberately left off** as pre-existing (not regressions you introduced): the post-login channel-tree sync race on the path lookup (same pattern as the shipped `autoJoinAfterLoginLocked`), and `rejoinLastChannelLocked` returning `true` before the join completes (established fire-and-forget convention). Worth a follow-up but not blockers for this PR.

Happy to pair on the state-lifecycle refactor if useful.

rfiorentino1:
All addressed — with one correction to the proposed fix that matters.

**The state-lifecycle split is in**, as you suggested: `stopReconnectTimerLocked()` extracted (timer only), and the success-path `cancelReconnectLocked()` moved to after the rejoin, which removed both the inlined teardown and the five snapshot locals. That also fixes the "stuck on reconnecting… forever" case, since a throw from `ensureOutputAudioReadyLocked` now falls into the retry path.

**But moving the cancel does not by itself fix the second drop.** `cancelReconnectLocked` still clears `reconnectPassword` at the end, and `connect()` remains its only other setter — so with just the reordering the feature still worked exactly once per session. A successful reconnect now explicitly re-arms the credentials for a future drop. Worth flagging in case the same assumption shows up elsewhere.

The rest:

- **Kick/ban** — both `MYSELF_KICKED` handlers now mark the logout as expected, and the drop handler honours it. A channel kick sends no logout, so the mark simply expires unused.
- **`connect()`** cancels any armed reconnect, so an A-timer can no longer tear down a live B session.
- **Instance leak** — the catch now closes the instance the attempt created; `destroyLocked()` was a no-op there because `self.instance` is only assigned on success.
- **Backoff** — one-shot timer re-armed after each failure, 5/10/30/60 s, giving up after eight attempts (~5 minutes) with "connection lost". Both numbers are arbitrary on my part; adjust freely.
- **Empty channel on `MYSELF_LOGGEDOUT`** — the channel is now recorded on every session publish while it is still valid, and the drop handler falls back to that.
- **Numeric-ID fallback** — verifies the ID still resolves to the same path before using it, and stores the channel password on that branch too.
- **Command-path double-signal** — `handleServerDropLocked` now reports whether it armed a reconnect; that case unwinds with an error the view controller stays quiet about, so no alert stacks on the reconnecting state.

**Cross-PR warning:** this branch and #37 each introduce their own `channelPathLocked(instance:channelID:)` in different files. No git conflict, but whichever merges second needs a one-line de-duplication.

math65:
Thanks — the core is right, and I confirmed the parts I had doubts about: `connect()` does cancel before it sets `reconnectPassword` (no ordering bug), the re-arming holds across N successive drops, the orphaned instance is closed on every failed attempt, and the counter does give 8 attempts before giving up. Your correction on the re-arming was the right call.

Rather than send you another round trip, I pushed the follow-ups straight to the branch — relegate or rework anything you disagree with.

**1. The "stay quiet while reconnecting" filter only reached 3 of ~19 command paths.** `presentError(_:)` is called from three sites; every other command failure in the connected screen goes through `presentActionError(_ message: String)`, which has already thrown the error type away. And since `.connectionLostReconnecting` deliberately shares its message with `.connectionFailed`, those paths could not filter on the text either — so a drop mid-command still stacked one modal alert per in-flight command on top of the reconnecting state, which is the exact thing the case was added to prevent. There are 19 callers of `waitForCommandCompletionLocked` across Administration / ChannelManagement / Messaging / Identity, so this was the common path, not the rare one. Added an `Error`-taking overload of `presentActionError` in both view controllers that have one, pointed the 20 call sites at it, and moved the test onto `TeamTalkConnectionError.isReconnectingDrop(_:)` so there is one predicate rather than three.

**2. `lastKnownChannelID` / `lastKnownChannelPath` were never cleared.** Neither `destroyLocked()` nor `connect()` reset them, so they survived into the next session: connect, join nothing (`TT_GetMyChannelID` is 0, so your fallback applies), take a drop, and the reconnect rejoins a channel from the *previous* session — resolved against the new server, so possibly a channel the user has never been in. Cleared in `destroyLocked()`, which is safe because `handleServerDropLocked` reads both fields before calling it. I deliberately did **not** also clear them when `publishSessionLocked` sees `myChannelID == 0`: a publish driven by another message can land after the SDK has wiped `m_mychannel` but before LOGGEDOUT is processed, which would reintroduce exactly the bug the field exists to fix. Within a session, keeping the last channel also matches what the preference is called.

**3. Both `MYSELF_KICKED` handlers marked the kick regardless of origin.** A channel kick sets the mark too, so a genuine `CON_LOST` within the mark's 5 s lifetime would suppress the reconnect and boot the user. `appendKickHistoryLocked` already draws the server/channel line with `message.nSource == 0` a few lines away — narrowed both handlers to that.

**4.** `handleServerDropLocked`'s doc comment had drifted above `reconnectBackoffSeconds` and was documenting the wrong declaration.

**Also, and this matters more than any of the above:** this branch was based on **v1.6.0, 108 commits behind main**. GitHub called it mergeable because nothing conflicts textually, but it had never seen 1.7.0, 1.8.0 or 1.9.0 — so you have been developing and testing against an app with no Channel Mixer and no `OutputAudioRenderEngine` (#22), and the `ttaccessibleTests` target does not even exist on that base. I merged main into the branch: it builds and the test suite passes, so no API drift — but your manual testing did not cover the current code, and that is worth knowing before the next branch. #35 and #37 are on the same 1.6.0 base; #33 is on 1.8.0.

PR #37, opened 2026-07-23
Persist channel passwords and skip the dialog when already known

rfiorentino1:
## Problem

Joining a password-protected channel **always** opened the password dialog, and the password you typed was remembered only in memory — gone after quitting the app or leaving/rejoining the server. So you re-entered it every session.

## Changes

- **Skip the dialog when the password is known.** A join now uses a known password directly (from this session, or saved from a previous one) and joins straight away. The dialog only appears when nothing is saved, or when a saved password is rejected (channel password changed) — in which case the new one you enter overwrites the stale one.
- **Durable per-channel passwords.** `ServerPasswordStore` now keeps a per-channel map in the keychain, keyed by the channel's **full path** (stable across restarts and channel-ID reassignment, unlike the numeric ID). A working password is auto-saved on a successful join.
- **Cleanup.** Removing a saved server now also clears its stored channel passwords, so no orphaned secrets are left behind. Old keychain items still decode fine (the new field is optional).

## Result

Enter a channel's password once → it's saved → future joins (this session or any later one, even after quitting or leaving/rejoining the server) join silently, no dialog.


math65:
Thanks — the feature is a real quality-of-life win and the `ServerPasswordStore` extension itself is clean (it reuses `loadCredentials`/`writeCredentials`/`nonEmpty` rather than duplicating keychain boilerplate). I ran a deep review and found a set of issues I'd want addressed before merge. The two biggest ones share a root cause, so I've grouped them.

## Root cause: no single source of truth for a channel password, and an unstable key

There are now **two** stores for "the password for this channel" — the in-memory `channelPasswords` dict (keyed by numeric channel ID) that the SDK-level join paths use, and the new keychain map (keyed by path) that only the manual VC join path uses. They never reconcile, and that split is behind the first two findings.

### 1. The persistence key is not stable — the opposite of the PR's goal

`channelPathKey` builds the key from `channel.pathComponents.joined(separator: "/")`, but the **root element of `pathComponents` is the server display name**, not a stable channel path. In `+SessionSnapshot.swift:155-156` the root channel's name is overridden to `serverDisplayName`, which is the fetched `szServerName` with a `record.name` fallback (`:104-105`). Two consequences:

- Within a single session the prefix can shift from `record.name` (before `TT_GetServerProperties` arrives) to the fetched `szServerName` — so a password saved early can't be read back later.
- If an admin renames the server, every saved per-channel password orphans at once.

Also the key drops the leading slash (`a/b`) while every other channel-path string in the codebase uses `/a/b` (`AppDelegate.swift:1043`/`:1785`, and `TT_GetChannelPath` output). I'd key off the SDK's own `TT_GetChannelPath` string instead — it's the stable, server-name-independent path.

### 2. Auto-join never consults the keychain, so "no dialog after restart" doesn't reach it

The new store is wired only into the manual VC `join()` path. `autoJoinAfterLoginLocked` (`+Connection.swift:270-341`) resolves its password from `options.initialChannelPassword`, the in-memory `channelPasswords` dict (empty on a cold launch, `:294`), and `connectedRecord.initialChannelPassword` — never `passwordStore`. So a user who saved a channel password via manual join, quit, and relaunches with join-last-channel enabled auto-joins with an empty password and lands in root, despite the saved secret. The fix for #1 and #2 is the same: one shared accessor, keyed by a stable path, that both the VC and the auto-join/reconnect paths call.

## Correctness / security — worth fixing

- **Repointing a saved server leaks channel passwords to the new host** (`SavedServersViewController.swift:604`). `editSelectedServer` keeps the same UUID (`makeRecord(id: record.id)`) and never clears the `channels` map, which is keyed by `id.uuidString`. If the user edits the entry to point at a different host, `knownChannelPassword` finds the old host's `channels["<path>"]` and auto-submits server A's channel password to server B via `TT_DoJoinChannelByID`, no prompt. Server removal clears the map; edit doesn't.
- **A rejected/rotated password is never invalidated** (`+ChannelActions.swift`, `.incorrectChannelPassword` branch). The keychain is written only on success; on rejection the code re-prompts (pre-filling the stale rejected value) but never deletes the entry. After a server-side rotation, every join silently re-submits the doomed password first and re-prompts with it pre-filled; if the user cancels, it's never cleared. A `setChannelPassword(nil, …)` on `.incorrectChannelPassword` would fix it.
- **Auto-save persists transient visits, with no way to forget** (`+ChannelActions.swift:401`). Any successful non-empty-password join writes to the keychain, including a one-off peek into a protected channel. There's no per-channel "forget password" UI — the only removal is deleting the whole saved server. Worth at least considering whether to persist only channels the user opts into.

## Smaller / follow-ups (not blockers)

- **Downgrade drops the map** (`ServerPasswordStore.swift`). `writeCredentials` re-encodes the whole `Credentials` with no schema version, so an older build that writes any credential for a server silently drops its `channels` map. Self-healing (re-prompt re-persists), but silent — a schema-version marker would make it explicit.
- **`ProfileDuplicator` doesn't copy the `channels` map** (`ProfileDuplicator.swift:163`) — duplicated profiles lose per-channel passwords. Fail-safe (dropped, not leaked). The `.tt` importer is fine; that format only has a single channel password.
- **Redundant keychain write on every known-password join** (`+ChannelActions.swift:401`) — the save is unconditional, so an auto-join rewrites the identical value. A guard helps, but note it must compare against the *persisted* keychain value, not `knownChannelPassword` (which already returns `password` by success time because `joinChannel` sets the in-memory dict first).
- **While you're in the store**: `baseQuery` sets no `kSecAttrAccessible`, so items default to `WhenUnlocked` (device-migratable). Pre-existing, not from this PR, but `…WhenUnlockedThisDeviceOnly` is the right call for per-device VoIP credentials.

Nice groundwork — the store extension is the right shape; it mostly needs the key made stable and both join layers pointed at one accessor. Happy to pair on the unification.

rfiorentino1:
All addressed. Your root-cause read was right — there were two stores and they never reconciled.

**One accessor.** The connection controller now owns both, behind `knownChannelPasswordLocked`, with `rememberChannelPasswordLocked` / `forgetChannelPasswordLocked` alongside. The view controller and the auto-join/reconnect paths all resolve through it.

- **Stable key** — now the SDK's own `TT_GetChannelPath`, not a path built from `pathComponents`. Exactly as you described: that one is rooted at the server display name, shifts from `record.name` to the fetched `szServerName` mid-session, and orphans everything on a rename. It also dropped the leading slash every other channel path carries.
- **Auto-join** — `autoJoinAfterLoginLocked` consults the store in all three branches. The cold-launch case ("join last channel" landing in root despite a saved password) is the one that actually mattered.

**Security**

- **Repointed server** — both editor paths clear the channel map when the host or TCP port changes. A rename alone keeps them.
- **Rejected password** — now forgotten rather than re-submitted and pre-filled forever. This also removes the pre-fill on the retry prompt, which is a change to pre-existing behaviour and a deliberate one: a password the server has just rejected should not be handed back to the user as the suggested value. Re-typing it is the point.

**On persisting transient visits:** I kept auto-save — it's the point of the feature — but added a way back out, a "Forget Saved Password" item on the channel context menu. It appears only when a password is actually stored for that channel and is hidden rather than dimmed otherwise. Menu validation can't afford keychain I/O, so the controller mirrors the persisted set in memory as channel IDs behind a lock: populated after login by resolving stored paths through `TT_GetChannelIDFromPath`, updated on every save/forget, cleared on teardown.

**Smaller** — schema version on the credentials blob (so a record last written by an older build is logged rather than mysterious); `ProfileDuplicator` copies the channel map; the redundant write is skipped by comparing against the *persisted* value; keychain items are added with `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (on add only — putting it in the lookup query would stop matching existing items).

**Cross-PR warning:** this branch and #36 each introduce their own `channelPathLocked(instance:channelID:)` in different files. No git conflict, but whichever merges second needs a one-line de-duplication.

math65:
Your root-cause read was right, and the unification holds up. I went through the merged result rather than the branch in isolation, and everything I asked for checks out: the key is now the SDK's own path; `autoJoinAfterLoginLocked` consults the store on all three branches; both editor paths clear the channel map when host or TCP port changes; a rejected password is forgotten rather than re-submitted; the schema version, the `ProfileDuplicator` copy, the redundant-write skip against the *persisted* value, and `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` are all in. Dropping the pre-fill on the retry prompt is the right call — a password the server just rejected should not be handed back as the suggestion.

I also checked the thing I was most suspicious of, the in-memory mirror of persisted channel IDs, and the design is sound: `markSavedChannelPassword` keeps it current on every save and forget, so it only goes stale for a channel that did not exist at login *and* already had a password in the keychain. That costs a missing menu item, nothing more.

Since #35 and #36 are merged, I pushed the follow-ups here rather than send them back to you.

**Rebased onto main — this branch was 118 commits behind, on v1.6.0.** GitHub reported CONFLICTING but that undersold it: it had never seen 1.7.0, 1.8.0 or 1.9.0, so you were building and testing against an app with no Channel Mixer and no `OutputAudioRenderEngine` (#22). One real conflict, in `+Connection.swift`, where your `refreshSavedChannelPasswordIDsLocked` call and main's `userVolumeStore` scope calls landed on the same line — resolved by keeping both, with the refresh immediately before the join that depends on it, on both the connect and the reconnect path.

**The stale base was also hiding a build break.** Adding a required `passwordStore` parameter to `TeamTalkConnectionController.init` breaks `DeviceStreamSourceTests`, which does not exist on the 1.6.0 base — so `xcodebuild test` could not have caught it on your side. Fixed, with the store pointed at a throwaway keychain service and defaults suite so the test cannot touch real credentials. Worth knowing generally: on that base, `xcodebuild test` fails with "scheme not configured for the test action", which reads like a broken setup but actually means the branch predates the test target.

**Deduplicated `channelPathLocked`** — #36 merged first, so the canonical copy now lives in `+Connection.swift` and this branch's identical one is gone, as you flagged.

**Finished the single-accessor unification.** Two older accessors survived: `rememberedChannelPassword(for:)` had no callers left at all once its one call site became `knownChannelPassword`, and `passwordForChannel(_:)` was still serving the .tt export off the in-session map only. The latter happened to be correct — every join populates that map first — but it is a second answer to the question this branch set out to give one answer to. Both removed, export repointed.

Two things I deliberately left alone, both yours to judge:

- `kSecAttrAccessible` is set on ADD only, so items written by earlier builds keep `WhenUnlocked` forever with no migration. You document it; I agree it is not worth a migration pass, but it does mean the hardening only reaches new items.
- The two `clearChannelPasswords` blocks in `SavedServersViewController` are byte-identical. Small enough that extracting it may not be worth the indirection.

rfiorentino1:
You flagged the stale base three times across #35, #36 and #37, and I want to address it directly rather than let it pass in the noise: that was my mistake, and I'm sorry for the extra work it made you.

I work on this with heavy AI assistance too, and this one got past me. The rebase was judged unnecessary on the grounds that nothing conflicted, and I never reviewed that call. "No textual conflict" says nothing about whether the work was built and tested against the current app — and it wasn't. On a 1.6.0 base there's no Channel Mixer and no `OutputAudioRenderEngine` underneath any of it, and the `ttaccessibleTests` target doesn't exist at all, so `xcodebuild test` couldn't have run on my side even in principle. The `passwordStore` break in `DeviceStreamSourceTests` is the concrete cost of that, and you're the one who found and fixed it.

Three things change going forward: every branch cut fresh from current `main`; the test suite run on the merged result before I push, not on the branch base; and when I've only exercised the happy path I say so, rather than calling it verified. That last one is what let the second-drop, kick/ban and connect-elsewhere cases in #36 through — I tested one drop and reported it as working.

One finding worth passing on, since it's part of the same problem: `xcodebuild test` fails with "Scheme ttaccessible is not currently configured for the test action" on current `main` too, not only on the old base. No scheme is tracked in the repo — `.gitignore` excludes `xcshareddata/xcschemes` — and the shared scheme's `TestAction` has no testables, so the only working one is a scheme Xcode auto-created in my `xcuserdata`. As it stands no contributor or CI can run the suite. Say the word and I'll send a shared scheme.

Thanks for pushing the follow-ups straight onto the branches rather than round-tripping them.

PR #38, opened 2026-07-27
Global mic-toggle hotkey: work while ttaccessible is focused, and use one chord everywhere

rfiorentino1:
The global mic-toggle hotkey became configurable in #31, but the focused case never followed it. `AppDelegate` installed the tap with `globalOnlyWhenInactive: true` — suppressing every press while ttaccessible was frontmost, on the assumption that "the menu shortcut covers the focused case" — while the User ▸ Microphone item stayed hardcoded at `.keyboardShortcut("a", modifiers: [.command, .shift])`. So once a user changed the binding, which is the whole point of it being configurable, nothing answered it with the app in front: the tap suppressed it and no menu item matched it. ⌘⇧A meanwhile kept working, which makes it read as random rather than broken.

Three commits.

**1. Defer to the menu only for a chord the menu actually carries.** A menu key equivalent reaches the focused app whatever a listen-only tap does, so acting on it as well would double-fire — but a chord no menu item owns has no other handler, and the tap should answer it. `mainMenuOwns(event)` compares what AppKit itself compares, `charactersIgnoringModifiers` against `keyEquivalent` with the hardware-only flags stripped, so letters, punctuation and F-keys follow one rule. It runs only after the chord has already matched, so ordinary typing never walks the menu.

Behaviour is unchanged where it already worked: ⌘⇧A focused still goes to the menu item, and a binding that collides with an existing in-app shortcut (⌘E, F5…) still fires only that shortcut. A pure-modifier chord can never be a menu key equivalent, so it no longer waits on a menu item that cannot exist.

Also requires a channel, matching the menu item's own disabled state — without it `activateVoiceTransmission` fails with `notInChannel` and pops an alert over whatever app the user is actually in.

**2. One chord everywhere once the hotkey is global.** The menu item now carries the configured binding, applied at the AppKit layer on the same re-apply cycle as the User menu's visibility (SwiftUI rebuilds the main menu and would restore its declared shortcut). ⌘⇧A remains the default and stays put while the hotkey is app-local only. `HotkeyBinding.menuKeyEquivalent` returns nil for a chord with no key-equivalent form, in which case the item gets no shortcut rather than being stranded on ⌘⇧A, and the tap owns that binding in every focus state.

**3. Name F13–F20.** They type nothing and carry no system function, which is what makes them the safe choice for a global binding — but the label table stopped at F12, so binding one showed as "Key 105" in the recorder and could not become a menu key equivalent either. The keys worth recommending were the keys the UI could not name.

Cut from current `main` (1.10.0). Builds clean; the suite passes with 7 new tests covering the ownership predicate in both directions and the binding → key-equivalent mapping, including the round trip between the two halves — they have to agree or the chord fires twice or not at all.

Note: this touches `HotkeyMonitor.handleTapEvent`, so it overlaps #34 textually. #34's version needs one extra term (`installedConsuming == false`), because a consuming tap swallows the chord before AppKit can route it to the menu and deferring would drop the press. Whichever lands second is a small resolution either way.

### DIFF
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -241,11 +241,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // stays muted forever in push-to-talk mode.
         pushToTalkMonitor.onPress = { [weak self] in self?.handlePushToTalkPress() }
         pushToTalkMonitor.onRelease = { [weak self] in self?.handlePushToTalkRelease() }
-        // The global ⌘⇧A tap fires only while another app is focused (the menu
-        // shortcut covers the focused case), so it mirrors the menu action —
-        // without restoring/focusing our window, which would steal focus.
+        // Mirrors the menu action without restoring/focusing our window, which
+        // would steal focus. The tap defers to the menu for a chord the menu
+        // carries (⌘⇧A by default), and answers any other binding itself,
+        // focused or not. Requires a channel for the same reason the menu item
+        // is disabled without one: transmission has no meaning outside a
+        // channel, and the SDK path would fail with "not in a channel" and pop
+        // an alert over whatever app the user is actually in.
         muteHotkeyMonitor.onPress = { [weak self] in
-            guard let self, self.menuState.mode == .connectedServer else { return }
+            guard let self,
+                  self.menuState.mode == .connectedServer,
+                  self.menuState.isInChannel else { return }
             self.connectedServerViewController?.performToggleMicrophoneShortcut(announceStatus: true)
         }
 
@@ -301,7 +307,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             muteHotkeyMonitor.configure(
                 binding: prefs.muteHotkeyBinding ?? HotkeyBinding.defaultMuteHotkey(),
                 scope: .global,
-                globalOnlyWhenInactive: true,
+                deferToMainMenuWhenActive: true,
                 wantsReleaseEvents: false
             )
         } else {
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -219,6 +219,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// owns that binding in every focus state, which is the same one-chord
     /// contract by a different route.
     ///
+    /// A binding that types (a bare or ⇧-only printable key) is withheld the same
+    /// way: an item carrying it would answer the key before a focused chat field
+    /// did, toggling the mic mid-sentence. The tap declines it while a field is
+    /// focused and answers it otherwise, so the chord still works everywhere it
+    /// isn't being typed.
+    ///
     /// Takes the preferences rather than reading the store: this also runs from
     /// the `$preferences` sink, where `@Published` fires in willSet and the
     /// stored property still holds the OLD value.
@@ -231,7 +237,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             return
         }
         let binding = preferences.muteHotkeyBinding ?? HotkeyBinding.defaultMuteHotkey()
-        if let equivalent = binding.menuKeyEquivalent {
+        if let equivalent = binding.safeMenuKeyEquivalent {
             item.keyEquivalent = equivalent.characters
             item.keyEquivalentModifierMask = equivalent.modifiers
         } else {
--- a/App/ttaccessible/AppDelegate.swift
+++ b/App/ttaccessible/AppDelegate.swift
@@ -205,6 +205,57 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             $0.submenu?.title == title || $0.title == title
         }) else { return }
         item.isHidden = menuState.mode != .connectedServer
+        applyMuteMenuShortcut(with: preferencesStore.preferences)
+    }
+
+    /// Keeps the User ▸ Microphone item's key equivalent equal to the global
+    /// mute binding, so one chord toggles the mic whether or not ttaccessible is
+    /// focused — the menu answers it here, the tap answers it everywhere else.
+    /// Declared in SwiftUI as ⌘⇧A, which is the right default and stays put
+    /// while the hotkey is app-local; a global binding overrides it.
+    ///
+    /// A chord with no key-equivalent form (a pure-modifier chord) leaves the
+    /// item without a shortcut rather than stranding ⌘⇧A on it — the tap then
+    /// owns that binding in every focus state, which is the same one-chord
+    /// contract by a different route.
+    ///
+    /// Takes the preferences rather than reading the store: this also runs from
+    /// the `$preferences` sink, where `@Published` fires in willSet and the
+    /// stored property still holds the OLD value.
+    private func applyMuteMenuShortcut(with preferences: AppPreferences) {
+        let title = L10n.text("shortcuts.microphone")
+        guard let item = findMainMenuItem(titled: title) else { return }
+        guard preferences.muteHotkeyGlobal else {
+            item.keyEquivalent = Self.defaultMuteMenuKeyEquivalent.characters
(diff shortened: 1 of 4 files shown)


math65:
Thanks — the diagnosis is right and it's a bug I hadn't spotted: once the binding is configurable, changing it makes the focused case go dead, and ⌘⇧A still working is exactly what makes it read as random rather than broken.

The approach is the right one too. `mainMenuOwns` compares what AppKit itself compares, so letters, punctuation and F-keys follow one rule instead of a table of special cases, and the round-trip test between the two halves is the test that had to exist — if they disagree the chord fires twice or not at all. Base is `e7a9a79`, current `main`, and the suite passes here (59 tests; the description says 7 new, there are 14).

One issue I'd want fixed before merge, plus a smaller one.

---

### 🔴 1. A binding with no modifier now toggles the mic while you type

`isValid` accepts any single key (`if keyCode != nil { return true }`, `HotkeyBinding.swift:39`), and the recorder reserves only Esc/Tab/Delete (`PushToTalkKeyRecorder.swift:99`). The model's own header offers "Space, ⌥Space, F5" as examples, and commit 3 actively encourages bare keys, since F13–F20 bind without a modifier. So a bare **K**, **Space** or **1** is reachable.

With the app frontmost and the caret in the chat field, pressing that key now:

- either hits the menu, which carries the bare key equivalent `applyMuteMenuShortcut` installed;
- or, if the menu doesn't carry it, falls through `mainMenuOwns` to the tap, which acts.

**Both branches toggle the mic mid-typing.** On `main` neither happened: `globalOnlyWhenInactive && NSApp.isActive` suppressed the press whenever we were focused. That suppression was also, incidentally, what protected text entry — dropping it is what opens this up.

This isn't an unconsidered case. The local path guards it explicitly, `HotkeyMonitor.swift:233`:

```swift
// A bare printable key (no modifiers) must keep typing into a
// focused text field — the field wins over push-to-talk there.
if binding.modifiers.isEmpty, Self.isPrintable(event), Self.isTextInputFocused() {
```

The global path never needed the equivalent, because it stayed quiet while focused.

**On the menu half, to be straight with you about what I actually established.** I tried to prove that a bare `keyEquivalent` steals the keystroke from a focused `NSTextView`. `menu.performKeyEquivalent(with:)` does return `true` with an NSTextView first responder and the window declining — which is precisely the call AppKit makes next. But I could not reproduce the capture end to end headlessly: neither `NSApp.sendEvent` nor a `CGEvent.postToPid` drives the key-equivalent path in a test binary, so I'm reporting the menu half as a strong risk rather than a confirmed one. The **tap** half needs no AppKit behaviour to hold — it's just the guard that's gone — so the finding stands either way. Worth noting the app currently has no menu shortcut without a modifier at all; this would be the first.

**Suggested fix**, both halves, or the problem just moves between them:

- in `applyMuteMenuShortcut`, install no key equivalent when the binding carries none of ⌘⌃⌥ **and** the key is printable — F13–F20, arrows and Esc stay eligible;
- in `handleTapEvent`, apply the existing `isPrintable` + `isTextInputFocused` guard when `NSApp.isActive`. Both helpers are already written a few lines up.

No test covers this today, which is why it got through a suite that otherwise covers the predicate well.

### 🟡 2. `mainMenuOwns` doesn't check whether the menu will actually answer

The predicate looks for an item that *carries* the chord, without consulting `isEnabled` or the application's state. The case that matters is an open `NSAlert`: the main menu is inert during a modal session, the tap defers to it, and the mic hotkey is dead — at arguably the moment when cutting your mic matters most.

Not a regression (on `main` the press was suppressed while active anyway), but this PR is the natural place for it, and it's one more term in the predicate: bail out when `NSApp.modalWindow != nil`, or test the matched item for `isEnabled`.

### Note, no action needed

With `muteHotkeyGlobal` off, the menu reverts to ⌘⇧A and a custom binding goes inert. Your description owns this, and it's defensible since the recorder is hidden in that mode (`PreferencesAudioView.swift:294`) — the UI stops promising anything. Flagging it only because the binding survives in prefs and comes back to life when the toggle is re-enabled.

### Checked and holding

- No tap/local double-fire: `scope: .global` installs only the tap.
- The tap callback runs on `CFRunLoopGetMain()`, so touching `NSApp.mainMenu` is main-thread-safe.
- Dropping the guard for pure-modifier chords is correct — no menu item can carry such a key equivalent.
- `applyMuteMenuShortcut` is reached from a `menuState.objectWillChange` sink with `.receive(on:)` and from `configureHotkeyMonitors(with: prefs)`, which is handed the value — the `@Published`/`willSet` trap is avoided on both routes.
- The menu item's `.disabled(...)` requires connection *and* a channel; the condition you added to the tap matches it.
- The ⌘⇧A default stays positionally correct on AZERTY.

---

Separately, on the shared scheme you raised in #37: you were right, and it wasn't only the stale base — `.gitignore` excluded `xcshareddata/xcschemes/` outright, so nothing was ever tracked and Xcode was auto-creating one per machine. A shared scheme with the test target as a testable is going onto `main`, so don't spend time on it.

rfiorentino1:
Both fixed — `2633b8b` and `21ae6e0`.

### 1. The typing case

The diagnosis is right, and the part I got wrong is that the `globalOnlyWhenInactive && NSApp.isActive` suppression was doing two jobs. I removed it for the one I meant to and didn't account for the other.

Fixed on both halves, since fixing one only moves the problem to the other. There is now a single predicate, `HotkeyBinding.typesIntoTextFields`, with three consumers:

- the menu installs `safeMenuKeyEquivalent`, which is nil for such a binding, so no key equivalent goes onto the item at all;
- `handleTapEvent` declines the press when `NSApp.isActive` and a field is focused — and only while we are active, since acting whatever has focus elsewhere is the point of a global binding;
- the local monitor consults the same predicate instead of its own `isPrintable` check, which is gone.

One predicate rather than two guards that happen to agree, because two would eventually stop agreeing and the failure is silent in both directions.

On the menu half: I couldn't reproduce the capture end to end either, for the same reason you couldn't. The item is withheld on the strength of your `performKeyEquivalent` result rather than a demonstrated steal. That costs nothing if it turns out the menu wouldn't have taken it — a withheld equivalent just means the tap answers the chord instead of the menu, which is the same one-chord contract by the route the pure-modifier case already takes.

### 2. I went wider than "printable", and it changes shipped behaviour

Writing the fix I couldn't defend the boundary. `isPrintable` is false for Return, and a bare Return binding is reachable — the recorder reserves Esc, Tab and Delete only when `mods.isEmpty` — so Return with the caret in the chat field would toggle the mic instead of sending the message. Same for the arrows, and for ⇧Tab and ⇧Delete, which that reservation doesn't cover.

So the predicate now asks whether a focused field would consume the key at all: printable characters, plus Return, Enter, Tab, Delete, Fwd Delete, the arrows and Home/End/Page by key code. F1–F20 are deliberately out — they do nothing in a text field, which is exactly what makes them the one family worth binding bare.

Two consequences to weigh rather than take from me:

- ⇧ is no longer treated as protective. The local path previously required `modifiers.isEmpty`, so a ⇧K push-to-talk binding swallowed the key while typing; it now yields to the field. That is a change to behaviour that shipped in #31, not only to the new global path.
- A binding in that family carries no menu shortcut, the same as a pure-modifier chord.

### 3. The modal case

`mainMenuOwns` now bails on `NSApp.modalWindow != nil`. I chose that over testing the matched item's `isEnabled`: with an auto-enabling menu, `isEnabled` is only refreshed while the menu is being displayed, so an item whose menu has never been opened reads stale — the check would look healthy in exactly the case that matters, where no menu is up.

### Tests

14 new; the hotkey suite goes 14 → 28 and the full suite passes at 73. They cover the predicate (each of ⌘/⌃/⌥ on its own, ⇧ not counting, the key-code family, F13 and F5 staying out), both halves withholding, and the two agreeing — for a typing binding `safeMenuKeyEquivalent` is nil *and* the ownership check reports the chord unowned, which is what routes it to the tap, where the focused-field guard then declines it. Neither half acts.

What they don't cover is the end-to-end keystroke, for the reason above.

Since this touches `handleTapEvent` again, the same guard needs to land on #34's copy of it; I'll carry it across if that one moves.

And noted on the shared scheme — I'll leave it alone.

PR #40, opened 2026-07-27
Build and test every pull request on GitHub Actions

math65:
Nothing verifies a contribution automatically today. The only check that has ever run on this repo is a maintainer building and testing by hand, on the one machine set up for it — and until `4cf99a9` a contributor could not run the suite at all, since no shared scheme was tracked (raised in #37).

This adds one job on `macos-latest`: fetch the TeamTalk SDK, build, run the unit tests. On every pull request, and on pushes to `main`.

**Scope, deliberately narrow**

Out: Developer ID signing, notarization, releases. A fork PR has no access to repository secrets, and `build.sh` already owns that path — duplicating it here would put a working chain at risk for no gain. Out too: any linter, since the project has none configured and it would only add red noise to otherwise sound PRs.

**Signing**

Ad-hoc (`CODE_SIGN_IDENTITY="-"`) rather than no signing at all: the app is sandboxed, and an unsigned bundle can carry no entitlements and will not launch on Apple Silicon — which the test host has to do. Verified locally with the exact overrides this workflow uses: `flags=0x2(adhoc)`, `TeamIdentifier=not set`, suite green.

**The SDK**

`libTeamTalk5.dylib` is git-ignored, so it is fetched per runner via `scripts/download-sdk.sh` and cached on a key covering the SDK version and both binary patchers — a version bump or a patch change invalidates it. `download-sdk.sh` prompts before overwriting, and a prompt with no stdin would abort under `set -e`, so it is skipped entirely on a cache hit; the build-phase patchers re-verify either way.

**What this PR exists to find out**

The test target runs against a `TEST_HOST`, so the suite genuinely launches tt-Accessible — on a runner with no audio hardware. The PortAudio probe patch should carry us through, since it removes exactly the device enumeration that would otherwise stall at startup, but that can only be confirmed by running it there. This PR is its own test case: if the checks below are green, the answer is yes.

Refs #37


math65:
Closing without merging — the CI itself is being set aside for now, but the two defects it exposed are worth keeping and have gone straight to `main`:

- `1608136` — the test target still carried Xcode's default `MACOSX_DEPLOYMENT_TARGET = 26.5` while the app targets 12.0, so the suite could only ever run on a Mac at 26.5 or newer. Together with the untracked scheme fixed in `4cf99a9`, that was the second reason a contributor could not run the tests at all (#37).
- `1c8ecd8` — `DeviceStreamSourceTests` guarded on an empty device list, but a machine can enumerate an input device whose IO server never starts. The guard now keys on whether the capture actually opens.

Neither was visible from the maintainer's machine by construction: the wrong deployment target is invisible on macOS 26.5, and the incomplete guard is invisible on a Mac with working audio.

The branch `ci/github-actions` is left in place if this is ever picked up again. What it proved: the build works on a runner with ad-hoc signing and no Developer ID, the SDK download and both binary patches run fine, and the suite passes 44/45 with the hardware-dependent check skipping cleanly. The open cost was runtime — `AudioDeviceStreamSource.start()` spends about nine minutes failing on a device that enumerates but never starts, which pushed each run to roughly twelve minutes.

